PR #32, opened 2023-06-27
fix typo, add auto-generated names, export svg

Omay238:
one is actually important, the other is just in the docs


Omay238:
@leomcelroy just fixed the second typo.

PR #38, opened 2023-06-30
make the examples work

Omay238:


Omay238:
I don't know why vercel failed. Maybe it was my changes in package.json?

PR #49, opened 2023-07-17
New editor implementation

Merlin04:
This PR (currently WIP) replaces the existing editor with a brand new version built with Astro, Preact, and Turbo.


Merlin04:
this moves everything to a monorepo so all vercel deployments will break - we'll probably have to fix once we merge

Merlin04:
we're at feature parity (plus more ofc)!

PR #127, opened 2023-09-24
cybertree-harshit

TGH75972:
This was really a great experience for me and I am glad that I perfected the turtle module's use along with NodeJS and I was able to create this artwork with the help of it, I felt like an artist when I gave this art a final touch.


leomcelroy:
Thanks for your submission! We haven't actually released Blot yet so your getting an early start.

The submission criteria is still evolving. We want people to have intent when creating their work so we're going to ask for figurative art (drawing inspiration from a real thing). I think your piece does that.

We also would like each artwork program to be capable of generating more than one instance. So you can change a parameter to make more in the "family" of the piece. To show this please submit at least three screenshots. I think your artwork could be adjusted to do that too.

If you add two more screenshots of your piece I'm happy to merge. Great to see this, thanks again. I think it's quite nice.

TGH75972:
I have added both the screenshots as you asked

leomcelroy:
Upon closer inspection it's evident that this was AI generated without much modification. Considering that I'm going to close this PR.

PR #129, opened 2023-09-24
slope_art-Soumay

BLADE9078:
It was very bewildering experience of writing as if it was like an artist.
My artwork describes the art made by straight lines and slopes between the by increasing the number of lines we can easily manipulate the art


leomcelroy:
Thanks so much for the submission! We haven't officially released the project yet so we're still working out the submission criteria.

You're piece is cool looking but we're hoping for people to have some intent when creating their work (as opposed to just running a program and seeing what cool thing comes out). 

Consequently we're going to try asking people to **make figurative artworks**. That means you should derive inspiration from real life things. 

Additionally the artwork needs to be parameterized in such a way that it can make multiple versions of the piece. Please submit 3 images.

Finally the piece must be drawable on the machine. That means it should comfortably fit in the blue box and not have too many lines or overlapping pieces (which correspond to pen strokes).

Thanks for your enthusiasm and helping us figure out the submission process.

PR #141, opened 2023-09-28
mosaic_art-Ansh

ansh3108:
I spent a major part brainstorming on what to make after, then I then searched internet and i found an interesting type of art- Mosaic art that is used in decoration of walls, floors and ceilings in historical buildings or temples. Plus this was quite simple too.


leomcelroy:
Thanks for the PR! This is a bit too similar to the starting examples though.

Try to make something representational/figurative as opposed to a abstract piece. Abstract art can also be great but for getting the Blot we want to see people are intentional with their work.

PR #144, opened 2023-10-01
mandalaSquares-innovativecreations

innovativecreations:
The code builds a mandala (kinda) using squares.


leomcelroy:
We're requesting people try to make representational/figurative art for their submissions (as opposed to abstract). 

That means you should try to draw something recognizable. Abstract art is great but for getting the Blot we want to encourage people to be intentional with the piece.

Please reach out if you'd like more guidance. Slack would be best if you are a teen and are on there.

PR #175, opened 2023-11-16
Fixes feature/bug where when the editor opens, the preview doesn't center where the drawing should be.

BrightTheBackpack:


leomcelroy:
Nice! Thank you.

PR #177, opened 2023-11-17
Account system

jianmin-chen:
Modals haven't been fixed yet, but everything else is functional. Combines the latest commit in the `main` branch with my changes inside `save-dev`.

Consider taking a look at `save-dev` because I think the other changes I've made in that branch are for the better. The file structure is better organized and I think is generally more understandable, all pages with the exception of the landing page are completed and working on mobile (although IIRC, the editor might need more work), it doesn't make use of Tailwind (which I personally - maybe more of a shared opinion - don't think is needed), and can completely builds and works (as demonstrated by the demo link that I shared).


leomcelroy:
I get this error when I try to run it locally.

Screenshot 2023-11-19 at 1 07 22 PM

jianmin-chen:
Mentioned issue should be fixed.

jianmin-chen:
Modals should work now for the most part. (You can see the share modal, and the save to email/remix modal is responsive to user input.)

PR #181, opened 2023-11-19
BitMap-AdityaPV

Vipremigini:
### Draw images through Blot
It is an implementation of BitMap imges on BLOT.
This code enables you to draw images with blot.

Steps to use:
1) Convert your image to binary using - [https://www.dcode.fr/binary-image](url)
2) Copy the binary and paste it in a .txt file
3) Access the .txt file with this code - [https://github.com/Vipremigini/Myfiles/blob/main/pixler.py](url) (This code will print a list, copy that)
4) Paste the list in index.js array as "pixl"

### Examples
steve
steve jobs 

logo
hc logo


V205Arduino:
Interesting.
I though Blot has a feature for SVGs?

leomcelroy:
I like this idea but could you include the program to turn the image into a blotable piece in the script you write. Right now the heavy lifting is done externally. I see your input as an image which is ascii encoded and the core of your program as the python script. 

Did you also write that python script? It seems so. Just port it over to JS and you're good to go!

leomcelroy:
Going to close this for now. Looking forward to seeing the updated submission.

PR #182, opened 2023-11-21
Fix missing variable declaration in leaf.md

khrj:
`y` is not declared previously and produces an error at runtime

image


leomcelroy:
Thank you!

PR #197, opened 2023-12-04
QR-sam

SamDev-7:
QR codes have always felt like magic to me. Over the past few weekends, I've been investigating how QR codes work and trying to implement my own code to generate QR codes. Along with many online resources (linked in the code), Blot was super helpful for me to visualize what was being generated without having to deal with complex graphics libraries. 

This generator supports all versions of QR codes, all error correction levels, and Unicode characters. 
Physical drawings are kept in mind, the line width and qr code size is configurable to be adaptable to all pen/marker sizes.

The code isn't as clean as I'd like, but I figured it was still worth submitting.
Let me know if cleanup is nesscary. 

Here are some examples!
> 
> 
> QR code with encoded data: "👋🏼 Hello from Sam's Blot QR Code generator! I ❤️ blot + hack club!"

> 
> 
> 
> Configurable line widths and piece sizes for ease of physical drawing

> 
> 
> Massive QR code containing fancy characters


leomcelroy:
Not exactly what we had in mind with programmatic art but it's for sure a fun visual project. Seems like you learned a lot and could be fun to add to other projects. Nice work!

You can [order your Blot here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

We're still sorting out shipping logistics so be patient but we're really excited to get the machine to you and hear your experiences with it.

PR #198, opened 2023-12-05
Add auth helpers

jianmin-chen:
Add auth helpers - four files in `/astro/src/db`. Also remove the old placeholder `/astro/src/database` folder as well as the `astro/src/pages/api` folder.


leomcelroy:
Looks good, thank you!

PR #199, opened 2023-12-05
feat: add self_portrait-kieran

kcoderhtml:
My art work is a self-portrait, using the SVG import feature to import an image of my face and then resampling and distorting it to make it look interesting. I also added shading around the SVG to make it visually more interesting and more of a challenge.

I have moderate experience coding with various languages for ~3-4 years and quite enjoyed using the editor except for the occasional crash whenever I made the code too complex. One useful feature would be a VS Code plugin that could deploy your file with live reload on a dev server hosted locally.

9b80b95741ddabee3b5fc979616ee3ca
5321a3518d259e067715d84f682346bd


kcoderhtml:
Thank you!

PR #203, opened 2023-12-16
sunflower-nguyengiabach1201

nguyengiabach1201:
sunflower-snapshot

It is a very relaxing and enjoyable moment when I create my masterpiece in Blot! Good job, Hack Club!


leomcelroy:
Thanks for sharing what you're working on! 

I think this is very similar to some example files. Try to think of some new concept you really like for a piece. Feel free to bounce ideas around on the Hack Club Slack #blot channel.

PR #205, opened 2023-12-17
arduinoPinout-yutaro

yutarour:
## Artwork description:
- I often write code for devices such as Arduinos, EP32, Raspberry Pi, etc. However, every time I do, I always have to go looking for the pinout diagram to see which pin is connected where. I thought it would be interesting to have the plotter be able to plot a Arduino pin out diagram. 

## Coding experience:
- I won't say I'm really good at coding (especially javascript and turtles). However, I do have some experience with embedded system programming (C++), and python.  However, I do have experience creating vector graphics in Adobe illustrate so that is what I decided to do. Fitting all of the labels for the pico proved quite a little bit of pain. 

## Images:
nano
uno
pico


leomcelroy:
I love the aesthetic of these! Is there some way you can make the art computational? 

One thing you could try is automating the labeling of different pins. Like in the image below. 

Better yet would be making it artistic in some way. How could you use the circuit in a composition?

Screenshot 2023-12-18 at 2 19 57 PM

yutarour:
Well the automating the pin labeling would be quite a pain as the pins are all in a different location

PR #208, opened 2023-12-18
collatz_conjecture-yutaro

yutarour:
## Artwork description:
- ### Why?
- The previous pull didn't go through because it wasn't code-y enough and used excessive amounts of SVGs. This time, I have created something that can be deemed as more "mathematical" or "artsy" (because in my opinion, the image looks like a wheat bushel). 

- ### What is the Collatz conjecture?
- The Collatz conjecture is one of the most famous unsolved problems in mathematics. When hearing that it is a "unsolved" problem, one might think that it is a very hard problem to solve, that involves rocket science. However, the Collatz conjecture unlike many other unsolved problems is actually quite a simple one to visualize (although it hasn't been proven **yet**) The conjecture is named after the mathematician Lothar Collatz, who introduced this idea in 1937. This problem has many names. such as the "3n+1 problem", "3x+1 problem", the Ulam conjecture, Kakutani's problem, etc. This problem says: "for even numbers, divide by 2", "for odd numbers multiply by 3 and add 1". When enough repetitions of this are done, will all positive numbers converge at 1?

## Coding experience:
- I won't say I'm really good at coding (especially javascript and turtles). However, I do have some experience with embedded system programming (C++), and python. I also have a particular fondness of numerical visualizations.

### it=200
collatz0

### it=500
collatz1

### it=1000
collatz2


leomcelroy:
Thanks for the resubmit! I think this may be a bit too mathematical. You mention the imagery of a wheat bushel. How about drawing one like this and maybe integrating the collatz drawing with it?

Screenshot 2023-12-19 at 10 57 36 AM

PR #211, opened 2023-12-19
Add yarn lockfile to git

maxwofford:
Fixes #209

PR #213, opened 2023-12-19
Cleanup gitignore

maxwofford:
This is an optional PR that goes a step farther than https://github.com/hackclub/blot/pull/211 and just removes all the duplications across the .gitignore files. Gitignores cascade, meaning lines in the top level gitignore `node_modules` and `astro/node_modules` + the astro gitignore of `node_modules` end up being redundant. This commit only changes lines in the gitignore that are already covered by another rule and shouldn't affect what's being gitignored as far as I can tell.


maxwofford:
The build is failing at the time of writing because #210 isn't merged into main. This should probably be merged after https://github.com/hackclub/blot/pull/211

PR #216, opened 2023-12-19
Don't stop execution if error position function fails

### LINKED ISSUE #214: console.{log|warn|error} fails on non-chrome browsers
in safari, console.log will throw and unhandled promise rejection.

maxwofford:
This is to address https://github.com/hackclub/blot/issues/214. As mentioned in this comment https://github.com/hackclub/blot/issues/214#issuecomment-1862162643, this PR doesn't fully fix #214. Rather, it adds a sane default to not crash the page's execution if `getPosFromError` fails to find where an error is.

### DIFF
--- a/astro/src/lib/run.ts
+++ b/astro/src/lib/run.ts
@@ -190,13 +190,20 @@ async function runCodeInner(str, globalScope) {
 }
 
 function getPosFromErr(err) {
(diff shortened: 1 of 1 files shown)


maxwofford:
This build will fail until #210 is merged

PR #220, opened 2023-12-20
my_OG_dedpool sketch

Samarpan1122:
i make this sketch in sketchbook app and then imported it to this blot code this drawing took me 3 hrs completely original
image


Hugoyhu:
Hi Samarpan! Very nice drawing. I don't necessarily think importing an SVG file into blot is quite the approach we're hoping for- could you try to take advantage of blot's built-in functions to make some more generative art?

PR #224, opened 2023-12-22
Emoji Generator

Samarpan1122:
Generate random emoji everytime

Screenshot 2023-12-22 131927 - Copy
Screenshot 2023-12-22 131912 - Copy


leomcelroy:
Some of these are a bit chaotic. Maybe being more transformative with the primitives could help, as opposed to just using plain arcs.

PR #227, opened 2023-12-26
Winter--Shrey

ShreyShingala:
So I wanted to make a Winter themed piece of art and I made this little scene. It features a cloud and has randomly appearing snowflakes. I was inspired to make it after experiencing my first green Christmas, during which there was no snowfall. Making the piece of art was actually pretty difficult/annoying. I first needed to find out to get math functions like Sine in Blot. After a bit of searching I figured out how to and started put off the hard task by making a cloud. Actually using the arcs was very difficult, but I realized  near the end you can drag the number to see how it would change in real time. This helped me finish the last bit of the cloud. I then actually made the snow flakes. Trying to figure out how to scale the snow flakes was the hardest bit. This it because I tried to use the inbuilt scale function, which was acting up. After lots of thinking and trying random things I tried to scale it without using the scale function, which finally did work. Overall actually making the art work was pretty fun. Wrapping my head around how Blot worked was pretty hard at first, as it was kind of like a learning curve, and I felt like giving up whenever something didn't work. However, I finally got the hang of it and powered through to make this.


leomcelroy:
I like this! Thanks for sharing. Could you make the cloud generative as well? This could be a good chance to look into noise. I think if the cloud was also generative you've earned your Blot.

PR #230, opened 2024-01-01
mazeGenerator-Evan

ligerbot:
Hello there!

For my blot generative art project I decided to make a maze generator for blot. I came up with the algorithm myself a while back and it consists of 4 basic steps:
1. Break wall in a random direction that is not been visited before
2. Mark current cell as visited
3. Move there (recursive function)
4. Backtrack if there are no open(not visited) places and repeat until no open paths are available

Thats it, seems pretty simple right? The interesting part I found out as I started the project is that it was the easier part, but it was only ~1/4 of the code. Writing all the support code such as the maze structure and making Blot be able to draw it efficiently was quite the fun challenge.

My previous coding experience lies in a smattering of languages, first scratch if you consider it a language and then learning c++ for Arduino in school last year. Since then I taught myself swift, the language taught in my school and then took a CS class. For this project I learned TypeScript & node.js and spent 10+ hours over 3 days on it.

My generator can generate a maze at any X/Y location on the paper and you are able to specify how many rows/columns you would like and a customizable set width/hight. 

Here is a little GIF demo (the quality degraded when converting to a gif)

Mazes

Here I generated a 10x10 maze that had a 2:1 hight to width ratio

squished_10x10_maze

Here is an example of the largest maze I was able to generate. Its 175x175 cells

175x175_maze-unrealisticForBlot

In the future I plan on trying new maze algorithms such as the ones I found on [this repository](https://github.com/shabbychef/mazealls). 

After I got everything working I tried switching out the random number generator for different sequences the first one I tried was a basic function that incremented by 1 each time

1-4 repeat

Another was up and down that incremented 1-4 and 4-1

1-4 up and down


leomcelroy:
Nice I can tell you put some thought into this and it brings up some really interesting ideas. 

I'd love to see it incorporated into a more artistic piece in the future.

If you're a teen you've earned your Blot! You can order it with this form [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #231, opened 2024-01-03
added lines+circles

yutarour:
## Artwork description:
- ### Why?
- The pull before the previous pull (#205) didn't go through because it wasn't code-y enough and used excessive amounts of SVGs. The last pull (#208) didn't go through because it was too mathematical. This is why this time, I did something that is artistic while having some math behind it. 

## Coding experience:
- I won't say I'm really good at coding (especially javascript and turtles). However, I do have some experience with embedded system programming (C++), and python. I also have a particular fondness of numerical visualizations (like this time).

## Images (all obtained by varying c):
s1
s2
s3
s4

Hope this pull goes better than the previous two because it's the new year and things are about to get busy again.


leomcelroy:
Nice work! Good job being persistent with this and I appreciate how you took three really different approaches. 

This code is very simple but produces an interesting visual result. Reminds me of those string art pieces.

If you're a teen, you've got your blot. You can [order it here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

yutarour:
@leomcelroy, Don't know if this is a bug or not (maybe a bug because the !metadata.json file is looking fine) but this project isn't loading from the gallery. 
image

I suspect the issue as underlined in the above image, is that the folder is missing the + (%2B) sign in the folder name and the .js file name. 

Actual URL:
https://raw.githubusercontent.com/hackclub/blot/main/art/lines%2Bcircles-yutaro/lines%2Bcircles.js

URL that blot is trying to load:
https://raw.githubusercontent.com/hackclub/blot/main/art/lines circles-yutaro/lines circles.js

leomcelroy:
Let's rename the folder.

PR #232, opened 2024-01-04
Generative Word Search - Scott C

Scott170c:
Our (me and my friend Lucas) "artwork" is a generative word search. You can control what words appear in it, and it generates something different every single time (even the word placement is different!). You can also change the size of it to make it harder or easier. While coding, starting was the hard part as we were figuring out js, and we were originally going to pull words from an api but decided against it for our sanity. After getting over that hurdle, it was fun programming together and seeing the project finally come together. Overall our programming experience was challenging but enjoyable!

snapshot02


leomcelroy:
This is fun! Also not what I had in mind when asking people to make art but I do like these drawn puzzles. It's a fun theme to come up and seems to be inspiring others. I like that you made use of another Blot project and extended it in an interesting way.

If you're a teen you can order your [Blot here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw). Congrats!

PR #233, opened 2024-01-04
Winter-Shrey

ShreyShingala:
This is the edited version of my original winter artwork. The main change that I made in this version is that I made the cloud generative and added noise. I used the noise to make the cloud change each time the code was run. The cloud itself is made up of three smaller cloud segments, which come together to make the large cloud. I made the cloud look like the way I did because I feel that it best represents what a cloud looks like right before a heavy rainfall/blizzard. The cloud itself can easily be adjusted as there are numerous variables to play around with.


leomcelroy:
The cloud is pretty chaotic now. One piece of criteria listed is that the drawing should be drawable. If there are a lot of overlapping lines the paper will tear. I was imagining something like this.

Screenshot 2024-01-05 at 5 03 58 PM

I've attached the adjusted code below.

I think some of the snowflakes could have a similar issue. Maybe you could adjust the snowflakes to look more like these.

Screenshot 2024-01-05 at 5 04 43 PM

```js
// DOCUMENT SIZE
const width = 300; // Width
const height = 200; // Height
setDocDimensions(width, height); // Setting height

// VARIABLES
const numPoints = 360; // Makes them smooth since there are 360 points
const increment = (2 * Math.PI) / numPoints; //incrament between points
const NumCloudPoints = 268; // Makes them smooth since there are 360 points

// CLOUD PART
const t = createTurtle(); // Making turtle
t.jump([50, 130]) //goes to first location

function cloud() {
  const t = createTurtle();

  t.arc(360, 27);
  t.scale([4, 1]);
  t.resample(1);
  const minX = t.lc[0];
  const maxX = t.rc[1];
  const centerY = t.rc[1];
  t.iteratePath((pt,t) => {
    const [x, y] = pt;
    const xT = (x-minX)/(maxX-minX) / 2
    pt[1] += noise([-0.9, xT*4.2])*22;
    pt[1] += (y < centerY ? -1 : 1) * bezierEasing(0.000, [0.626,0.381], [0.253,-0.026], 0.0000)(xT)*24;
  });

  t.translate([143, 173]);

  drawTurtles([ t ]);
}

cloud()

//SNOW PART
const snow = createTurtle();

// uses math to make points!!!
function generatePoints(theta, a) {
  const r = Math.sin((a / 5) * theta);
  const x = r * Math.cos(theta);
  const y = r * Math.sin(theta);
  return [x, y];
}

// put the turtle in the middle
snow.translate([width / 2, height / 2], t.cc);

const numSpirals = randIntInRange(10, 25); //How many snow flakes there will be

for (let i = 0; i < numSpirals; i++) {
  const randomX = randIntInRange(10, 280); //Random X
  const randomY = randIntInRange(10, 88); //Random Y
  const randomA = randIntInRange(15, 100); //Random type of snowflake
  const scaleFactor = randIntInRange(2, 10); //Makes it big/small

  //gets the center to the correct rand point
  snow.up();
  snow.goTo([randomX, randomY]);
  snow.down();


  for (let j = 0; j <= numPoints; j++) {
    const theta = j * increment; //increaments it
    const [x, y] = generatePoints(theta, randomA); //makes the point to go to

    // Manual scaling cause the scale factor was being weird
    const scaledX = x * scaleFactor;
    const scaledY = y * scaleFactor;

    snow.goTo([scaledX + randomX, scaledY + randomY]); //Goes to the point
  }
}


//DRAWING THE IMAGE
t.join(snow) //joins both together

drawTurtles([t]); // actually draw it
```

leomcelroy:
How could we make the clouds look more like these?

Screenshot 2024-01-05 at 5 07 13 PM

Screenshot 2024-01-05 at 5 08 41 PM

leomcelroy:
This is pretty simple to generate conceptually but we need to think about how the booleans would work.

Screenshot 2024-01-05 at 5 20 29 PM

V205Arduino:
What if you use bigger curves and longer straight lines?

ShreyShingala:
I'll get to work changing the snowflakes and making them look like the ones in the picture.

ShreyShingala:
I redid the snowflake system and added 5 distinct types of snowflakes. This should fix the issue of the paper ripping because of too many lines. 


Screenshot 2024-01-06 at 4 22 39 PM
Screenshot 2024-01-06 at 4 22 56 PM

leomcelroy:
These snowflakes look really good!

leomcelroy:
Is the PR updated. The files seem to be the same.

ShreyShingala:
Oops, it appears I forgot to do that

ShreyShingala:
I changed the code, but I'm not sure how to change the snapshots.

leomcelroy:
Apologies I was unclear as to whether you were waiting for me or the other way around. I think the snowflakes look great and the cloud is on it's way. Two more thoughts.

Make the flakes smaller and more numerous. Try to prevent them from overlapping with each other.

Give some variation to the cloud.

With these two changes you're there!

leomcelroy:
I'm going to close this for easier house keeping but please reopen or make a new PR when you are ready!

PR #235, opened 2024-01-06
Add files via upload

ameya-g-git:
basic sunset scene with a cityscape in front and waves along the bottom edge of the page. was very fun to figure out the math behind constructing each individual component (how to make it so that the sun doesn't go through a building, how to align the buildings side-by-side, generating a random wave pattern each time), and i'm pretty happy with the final result.

i've been programming for a while now, since maybe 3 years ago, and i'm currently really passionate about intersecting design and programming to create cool and unique user experiences. pretty proficient with python, c++, html, css, js, and a a bunch of libraries and frameworks within these languages. ill attach the sample images here in case they dont show on the PR
sample1
sample2
sample3


leomcelroy:
Nice I like that it's a composition. Good variety in the generated pieces too.

If you're a teen you've earned your Blot! You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

ameya-g-git:
thanks! looking forward to getting my blot!

PR #238, opened 2024-01-13
chakra-Riston Rodrigues

ristonrodrigues723:
It was inspired by the ashoka chakra, initially i had planned to do something entirely different (circles , a whole lot of them forming an artwork of concentric and intersecting circles  ) but it was not posssible to get the results as i anticipated, so had to change my art piece.  I included code from earlier pieces i was trying to derive (one including a mandala) and finally derived this piece after many changes and iterations. took a lot of time as i had to start from scratch and proceeded to make changes as i learned and kept on improving as needed. it was a mixture of a lot of fun , head scratching and a lot of thinking that helped me finally get to the current art piece.

blot3 (2)


leomcelroy:
Thanks for the submission. I think this piece is still just exploring the toolkit. Try to be more transformative with what the Blot Turtle can do. I recommend finding some example pieces which you like which are drawn by hand thinking about how to distill the art into a program.

ristonrodrigues723:
> Thanks for the submission. I think this piece is still just exploring the toolkit. Try to be more transformative with what the Blot Turtle can do. I recommend finding some example pieces which you like which are drawn by hand thinking about how to distill the art into a program.

hey bro my original 
image
picecd is this however it doesnt sem to be drawable through blot

leomcelroy:
That would be a great piece and I think all is doable. You'll have to think about how to make the shading lines.

How far did you get?

leomcelroy:
I'm going to close this for now but please re-open when you are ready!

ristonrodrigues723:
hey bro i managed to do this man this really made me scratch my head maybe more revisions are needed

ristonrodrigues723:
tried doing the overlapping but it hides the full circle seemingly and not the part that overlaps

PR #239, opened 2024-01-16
Dexter-sunnyhouse

Speedlet2:
This is a modified version of the code that I wrote (and then promptly lost) at Christmas of a house in the snow (whoops)! To represent the change in season, it now shows a house in the sun, with a pathway leading up to the front door. As with my previous artwork, this piece changes based on the aspect ratio put into the base 'width' and 'height' constants. As I've said in my previous PR's for blot and sprig, most of my experience with programming is with Python and C#, but both of these tools have started my journey to learning JS too! I hope this helps and that the artwork is more similar to what you are looking for!


leomcelroy:
Pretty short script but it draws a recognizable thing and is parametric. Nice work remaking it after losing it.

Congrats, you've earned your blot. 

[You can order it here.](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw)

PR #241, opened 2024-01-17
Emoji gen-samarpan

Samarpan1122:
emoji generator new pr with code optimised.


leomcelroy:
Definitely better but these are still very basic shapes. In the examples we've been passing back and forth there are curves that would require more than just one line or arc. Try to recreate something of the level of detail and polish as one of the ones below.
 
Screenshot 2024-01-27 at 10 52 48 PM

Samarpan1122:
made some changes

leomcelroy:
I ran it a bunch of times and it's quite charming to watch the faces transform. I think there are further places you could take this concept but also feel you've earned your blot.

Congrats! [If you're a teen you can order it here.](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw)

PR #255, opened 2024-01-23
Add initial blot packing instructions draft

maxwofford:


maxwofford:
Merging because I just used these to back my first blot and they worked!

PR #256, opened 2024-01-24
Fix the new arcing implenetation so that the final angle is correct.

BrightTheBackpack:
See https://hackclub.slack.com/archives/C04GCH8A91D/p1706065513565839?thread_ts=1706032016.476829&cid=C04GCH8A91D


leomcelroy:
Nice catch! Thank you.

PR #258, opened 2024-01-25
Editor page updated meta data

brunoblaise:
Editor page lacked def meta headers


leomcelroy:
Thank you!

What's the purpose of the highlight.js stylesheet?

brunoblaise:
Web capture_1-2-2024_153644_blot hackclub dev

It is used to create syntax highlighting as you see in the image I didn't add it I just found it in the head, it was being used for syntax highligthing on the editor page, I just made it global and it doesn't affect any thingon other page,  only on the editor page for toolkit and workshop.

PR #263, opened 2024-01-28
starwarsthing-birbexe

BirbEXE:
my "artwork" (if you can call it that) is the starwars logo surrounded by lines, designed to mimic hyperspace

This is actually only my second time ever coding in javascript, and my only successful time (I gave up first time around).
I am very fluent in python, and somewhat fluent (although a bit rusty) in html and css.

this thing took me an unreasonable amount of time to make

snapshot-1


leomcelroy:
I like the concept but this relies too much on importing an SVG. What are some Star Wars themed things which could be fun to generate algorithmically? Maybe a galaxy?

I'm going to close this for housekeeping, but please re-open or submit a new one when you are ready.

PR #268, opened 2024-02-01
add star--legion

AgentM7:
This art is a kind of star and sometimes it looks like a flower, you might have to run it a few times to get a good output, they are too dense sometimes. For coding i was first trying random things and trying out random things and testing the functions from toolkit and then actually thought of making something and ended up making this


leomcelroy:
Thanks for submitting! I think this is a bit minimal. Maybe you could work on the concept a bit and think about what you're aiming to make. Looking forward to a future PR but will close this for now to keep the triage manageable. Just to emphasis always welcome to submit again!

PR #277, opened 2024-02-04
Assembly page is astro-configured redirect

maxwofford:
This means the link is only in once place and less likely something we can screw up by partially updating


leomcelroy:
Thanks!

PR #278, opened 2024-02-05
Create TUNING.md

maxwofford:
_Something something you can't tune a fish..._

So you built your blot, what now? You need to tune it! This guide will be for first-time assembly as well as debugging if your blot starts drawing poorly.


BrightTheBackpack:
***My lot isn't drawing precisely***
This probably means the belt is loose. Try tightening the timing belt pulley by screwing it in more, or try tightening the belt by moving the belt clip forward, attaching the belt to the clip as tight as possible, and moving the belt clip back.

maxwofford:
@BrightTheBackpack I want the tuning guide to have very specific things to fix (ie. diagonal lines == only 1 motor is working). Would you be up to make your belt too lose and try drawing the blot logo so we can get video of the issue?

PR #282, opened 2024-02-05
Clouds-Suchit

Suchit-07:
The artwork are some random arcs that make a different, unique, random, cloud every time.
Screenshot 2024-02-05 180050


leomcelroy:
I like this concept and you're getting a good effect but I think you need to bound it in a cloud shape.

Like such.

Screenshot 2024-02-08 at 5 23 03 PM

leomcelroy:
I'm going to close this for now to keep house keeping easier but please reopen later!

PR #284, opened 2024-02-07
blossom-reva

revaga:
with this art piece I first wanted to make penrose tiles, so I started out with rhombuses (but the decision-making/math was too confusing, w/ very few resources)
I realized that the rhombuses looked like flowers, so I wanted to make something that looks like a pretty flower garden.
then I played around with them, turning them into rows of flowers (I got this really cool 3-d effect at one point!)
I decided to make rings of flowers and then once I created one ring, I played around with for loops to create multiple rings of different sizes and layouts.
overall, the process took me over 2 weeks (with some breaks in between, probably in total 8 hrs)

now, the art piece looks like blooming flowers, with the outermost flowers being the biggest and the innermost one the smallest (kind of like fireworks but flowerworks)


leomcelroy:
Nice work being persistent and thanks for the submission! I think this is still a bit short. Though not a perfect measure a minimal artwork would be around 100 lines. 

How far did you get with the Penrose tiling?

revaga:
Note: I added some more comments in my latest update

I didn't get very far because the penrose math I found required object oriented programming or complex algorithms (that were not working in blot for me).

Can you clarify what revisions you would like me to make?

revaga:
hey! @leomcelroy I made some revisions, is it better now?

leomcelroy:
I think it looks pretty nice in many generations but the placement of the shapes can be a bit haphazard with them overlapping. It's definitely possible to do a tiling and I would strongly encourage persisting with that. I can help out where you get stuck. Can you share an image of what your are aiming for?

revaga:
I updated the code so that they don't overlap as much, thanks for catching that! Is it better now?

revaga:
I've updated the code so it looks more like an organized garden rather than flower arrangement, is it better now?

revaga:
I'm not sure how to open another pull request, but I've uploaded code for my penrose tiles to this one. For the math/algorithm behind it, I looked at a book called L-system Fractals by Mishra which was a very cool read!

leomcelroy:
Nice! Can you show a few productions to demonstrate the sort of patterns this can make. With that this should be good to go.

Let's close this PR and you can open a new one the same way you opened this the first time, that way we won't also pull in the outdated files when we merge.

PR #287, opened 2024-02-08
Michelle Blot PR

michelle12349502039:
I had a lot of fun figuring out how to draw a spider web using blot. I thought of the idea based on a fractal tree gone wrong attempt so I figured I could just make a spider web out of it. When I was making it, I realized that there was a lot of math involved in drawing it which I learned in previous high school math classes that I thought I'd never apply but without them, I wouldn't be able to complete this drawing. I utilized distance formula and trig (to get a point on a circle and find of the angle between each radial line). A way to make this more sophisticated is to use bezier curves instead of lines for the horizontal lines of the spider web and I may attempt that next time. 
Here is an image of my artwork:
chrome_OObSYxS5Rz


leomcelroy:
It does look like a web! Do you think you can level up by adding this effect?

Screenshot 2024-02-08 at 5 11 31 PM

It's enough to earn your Blot though. [You can order it here.](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw)

PR #288, opened 2024-02-08
Improve packing & testing for rp2040

maxwofford:


leomcelroy:
The last thing to do with a firmware change is rebuild the uf2 file.

PR #292, opened 2024-02-13
Added Oskar's Mandelbrot generator.

admiralpav:
Added a Mandelbrot generator that randomly picks regions of interest on the Mandelbrot set, and renders them in two distinct methods.


leomcelroy:
Great work! I appreciate that you made the math inspired piece aesthetic as well.

You can [order your Blot here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

Issue #174, opened 2023-11-16
color contrast


LucasHT22:
Maybe it's just my perception but the contrast of blue and white can be uncomfortable in the long term

V205Arduino:
Perhaps a dark mode would be nice....

PR #305, opened 2024-02-17
Add Dark Mode

### LINKED ISSUE #194: Dark mode or customizable pallette
Know this seems extravagant, but I find I'm most creative at night and really want to use blot on a dark screen.

I'm 100% ok with not having official support for dark mode; but currently the way the turtle is rendered, extensions like "dark reader" will black out the canvas entirely:

Screenshot 2023-12-01 at 12 59 39

Short of a full dark-mode, I'd like the ability just to change the color of the pen in the renderer to work with the extension.

BrightTheBackpack:
closes #194 #174

### DIFF
--- a/astro/src/components/Console.module.css
+++ b/astro/src/components/Console.module.css
@@ -9,9 +9,12 @@
 }
 
 body:not([data-theme='dark']) .root {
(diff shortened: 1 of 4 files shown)


leomcelroy:
I think this needs to be updated for the non astro version.

PR #308, opened 2024-02-17
Add t.pos

### LINKED ISSUE #265: No position function outwards facing
Could we add an outwards facing position function to blot? Currently there is nothing that does this to my knowledge, and the position is already stored in the class. This should only take a few lines of code. 
I believe the function below could be added to [`virtual-gallery/drawing-functions/Turtle.js`](https://github.com/hackclub/blot/blob/main/virtual-gallery/drawing-functions/Turtle.js) to implement this functionality.
```javascript
pos() {
    return this.location
  }
```

BrightTheBackpack:
Closes #265 
I haven't done a lot of testing with this so maybe do testing before merge

### DIFF
--- a/astro/src/lib/drawingToolkit/turtle.ts
+++ b/astro/src/lib/drawingToolkit/turtle.ts
@@ -357,7 +357,9 @@ export class Turtle {
 
     return [pt[0], pt[1]]
   }
-
+get pos():  Point{
+    return this.location
+  }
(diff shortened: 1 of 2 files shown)


leomcelroy:
This also has tuning docs in it. Can you open a new PR with just the pos change.

BrightTheBackpack:
done!

PR #312, opened 2024-02-20
L-System Fractals-Reva

revaga:
I was exploring L-systems as a way to code phenomenon in nature as generative art, and came across Penrose tiles (the first image). It look me a while to figure out how to convert the axioms into blot from the L-system Fractals by Mishra book.


leomcelroy:
Okay almost. When I mentioned give a variety I meant to adjust the rules and production string.

This stuff.

```
var axiom = "[X]++[X]++[X]++[X]++[X]";
let ruleW = "YF++ZF----XF[-YF----WF]++";
let ruleX = "+YF--ZF[---WF--XF]+";
let ruleY = "-WF++XF[+++YF++ZF]-";
let ruleZ = "--YF++++WF[+ZF++++XF]--XF";

var production = "[+-WF++XF[+++YF++ZF]-F--ZF[---WF--XF]+]++[+-WF++XF[+++YF++ZF]-F--ZF[---WF--XF]+]++[+YF--ZF[---WF--XF]+]++[+-WF++XF[+++YF++ZF]-F--ZF[---WF--XF]+]++[+-WF++XF[+++YF++ZF]-F--ZF[---WF--XF]+]";
```

This can produce lots of different drawings. I wanted to see some samples from that space. Look up l-systems to get a sense of what can be done.

Also please follow the [folder format outlined here](https://github.com/hackclub/blot/blob/main/docs/GET_A_BLOT.md#22-place-your-program-in-a-folder-with-at-least-3-snapshots).

revaga:
My new drawing is a levy c curve (which kind of looks like a mushroom cap!)

revaga:
I updated the code for mutliple Lindenmayer systems including a dragon curve, Koch curve, and branch curve.

revaga:
Apologies for all the commits, I would love to have this submission reviewed again! Rather than focusing on a single Lindenmayer system, it is compatible with quite a few!

leomcelroy:
Very close! It seems that you initialize your array for storing position as a number then start pushing to it. This works sort of because of how loose JavaScript is with types but could also cause problems. Also you should be saving the angle too. Try to make an object with both position and angle and push them both to an array of saved states.

leomcelroy:
Please name the js file `index.js` and make sure no images have spaces in the name. Also the snapshot referenced in the JS has the wrong name.

revaga:
updated! index.js file, more images w/ no spaces in the file names

leomcelroy:
Great work persisting with this! Thank you for adding this classic idea to the gallery, I hope you learned a lot in the process.

Congrats on earning your Blot! You can [order it here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw). It may take a little while to receive because we have limited supplies and are still sorting out logistics. Looking forward to seeing your art work drawn.

PR #313, opened 2024-02-20
add t.pos

### LINKED ISSUE #265: No position function outwards facing

BrightTheBackpack:
closes #265

### DIFF
--- a/astro/src/lib/drawingToolkit/turtle.ts
+++ b/astro/src/lib/drawingToolkit/turtle.ts
@@ -357,7 +357,9 @@ export class Turtle {
 
     return [pt[0], pt[1]]
   }
-
+  get pos():  Point{
+    return this.location
+  }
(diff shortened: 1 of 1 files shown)


leomcelroy:
I think you may want to deep copy it by destructuring x and y and returning in new array. Someone may use the position to do something and inadvertently move the turtle.

leomcelroy:
Nice. Thank you.

PR #314, opened 2024-02-22
Layered Lotus - Charles Huang

Yoloholoknow:
My artwork depicts a flower that spirals such as a lotus. Using the variables, one can manipulate the number of layers the lotus has an the shading of the layers. My experience coding this was quite exciting because I am not that proficient in javascript, so learning turtle was quite a interesting task.
snapshot1


leomcelroy:
This is interesting but a bit too simple. How could you make it more sophisticated? Perhaps something like this.

Screenshot 2024-02-22 at 10 21 40 AM

I'll close this for housekeeping but please resubmit when ready.

PR #317, opened 2024-02-27
ParametricTorus-KoKoOo

KoKoOogit:
I have been interested in mathematical drawings for a long time. So, I made a torus from a circle equation and fit the drawing into the window (dimensions). I found some problems while using for  loops,  the program crashed if an increment is less than 1. I think the problem might be in platform.


leomcelroy:
Thanks for matching the submission format well! I like the idea of drawing a torus but think this piece isn't quite there with the approach. The basic structure of many "too simple" designs has been a basic geometric pattern (circle) repeated with a straightforward affine transform (translate, rotate). I think this fits that mold pretty well. One of the issues is the visual 3D effect only shows up when the iteration counts are very high. But these pieces with many overlapping lines won't draw well on a machine. How could you get the 3D effect with a different approach. 

What do you see in this image?

Screenshot 2024-02-28 at 11 40 24 AM

leomcelroy:
I'm going to close this PR for housekeeping but you can ask more questions here and open another one later for a new submission.

PR #321, opened 2024-03-01
lampda-snipchu

snipchu:
image
the artwork is a little lamp and his name is lampda!

i thought making this wouldn't be too hard, but i ended up taking around 4 hours to make it! i mostly spent my coding time adding parameters for True Customization (but i did have to put some restrictions for some parameters for physics purposes)

and making the lamp connect to the lampbox was really difficult (since my origin point started at the end of the arc, not the middle)!! but eventually i solved the problem with a concerningly long line of code


leomcelroy:
Very cute. You could make a seen where they all point different angles. Like a vector field with lights. Or raycast from the bulb onto something. I think it's Blot worthy but hope you keep developing it!

Congrats, [here is the order form](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw). We are sorting out some logistics so it may take a little while to get your Blot but you'll receive an email when it ships.

leomcelroy:
Oh I did notice that your snapshot name has to match the name in the metadata. Can you fix that?

PR #324, opened 2024-03-02
Minesweeper-LiterallyEverything

OHIO-MAN:
My artwork is a program that can randomly generate a minesweeper board. The parameters you can change are the mine density and the board size.
My experience coding was very fun and new for me as this is my first time coding in JavaScript.
Also here is a picture of my artwork:
image2


leomcelroy:
I would love to see you try something leaning a bit more towards the artistic but your execution of this is very good and the result comes out nice. I am wondering what it's like to play Minesweeper with pen and paper, the game may not translate so well. 

Overall good work. I think you've earned your Blot. If you're a teen you can fill out the [order form here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw). It may take a while to receive because we're still sorting out logistics and have limited supplies.

PR #325, opened 2024-03-04
TheFace-Ari

ari-kl:
snapshot0

I decided to make a face (+arms) with different parts that can move around. While it is visually simple, the face can still show multiple expressions with the random variation it has.

While I do have experience programming, I've usually done more non-visual things (for example involving networks) and want to get more into design-based projects.

Overall this took me a few hours to get right. To optimize drawing and abuse the symmetry in my work, I made a function that lets me rotate differently based on a given direction which lets me draw both sides equally with minimal changes.


leomcelroy:
Thank you for the submission. I like where you are going with this but think we could level up the piece some more. You mention working with networks. How about making a network geometric looking face?

Something like this?

Screenshot 2024-03-05 at 1 34 15 PM

ari-kl:
Thanks for the suggestion. I rewrote my code and created something similar (and more expressive than before).

The way I did made it is I created a base set of points where the points are then transformed within a certain range on both axes to create variation in the face. The right side is generated based on flipping the left-side points x-coordinates across the midpoint line and adjusting the connection ids. 

To keep it closer to a face structure, the transformation range is set to 3 (that mean a -3 to +3 change in position on each axis) but can be adjusted at the top of the file.

leomcelroy:
Thanks for your patience. It seems to have lost some of the face qualities. How could you make the features more recognizable?

I've been working on incorporating feedback into the updated toolkit. It is described in the editor toolkit on the editor page.

Would love your feedback after using both. Let me know if you have any questions!

I'm going to close this for housekeeping but please resubmit! You've got a good concept.

PR #328, opened 2024-03-07
a few small bugfixes

rivques:
- Allow CRLFs in the frontmatter of guides via a regex tweak in extractFrontmatter.js (yes, i was on Windows, why do you ask?)
- Make the default program use the drawLines function instead of the nonexistent draw function
- Fix a bug in Turtle.forward that made it break when used after a pen-up jump


leomcelroy:
Thank you! Nice catches.

PR #331, opened 2024-03-10
Graphing-Calculator-stati30241

stati30241:
My artwork is a graphing calculator with which you can graph many different types of curves. Not only can you graph functions, but also implicit curves, which are much harder to graph. I put some examples of functions in the code, but you can create your own and it will graph it. You can also change the window in which the curve is graphed, and the resolution at which the graph gets drawn. The drawing of the curve is done using the marching squares algorithm, which a very interesting algorithm.

I had plenty of fun coding this. I had implemented marching squares once before in C++, but I had little experience in javascript, and none with Blot. It took me a while and a dozen google searches, but I got it working eventually.

x^2 + (y - sqrt(abs(x)))^2
sin(x + y) - cos(xy) + 1 = 0


leomcelroy:
I like the idea, it's a bit of a tech demo at the moment! Could you make a proper composition using this?

stati30241:
Can you be more specific as in what you mean by a proper composition? Would graphing something like the batman equation with my calculator be enough?
image

leomcelroy:
Check out this video to get a sense of using equations to make something really artistic.

https://www.youtube.com/watch?v=8--5LwHRhjk

I think you're almost there you have a great technique in hand!

PR #332, opened 2024-03-10
pi-visualisation-selo

Han-Selo:
### Visualisation of Pi

This project is based on this [article](https://www.scientificamerican.com/blog/sa-visual/the-boundless-beauty-of-pi/) by Amanda Montañez:

At the top of the index.js file, the variable "lines" can be altered to produce different snapshots.

The result is very similar to images of [enzymes](https://news.mit.edu/2018/scientists-deliver-high-resolution-glimpse-enzyme-structure-0220) , which I find to be quite interesting, another instance of math in nature I guess!

Overall, I had a very smooth experience with the Blot editor 👍

Here's a snapshot of a possible visualisation (the first 10000 digits of Pi in this case):
10000_digits


leomcelroy:
I like the concept a lot but think the result isn't as beautiful as the piece shown in the blog being monochromatic instead of colorful. Could you think of your own Pi visualization? 

PR #333, opened 2024-03-12
subsected-rayhanadev

rayhanadev:
Generated snapshot of Ray's Blot project, Subsected. It has a maze-like design with additional subsections in the plot.

Subsected is a generated art piece that uses a recursive geometric generative algorithm that subdivides squares, with some added spice, to create interesting geometric forms. It generates additional "subsections" in the pathways, enhancing the final product!

You can alter the number of iterations and the size of each subsection to create new designs! These figures primarily affect the density of the final output:

Another generated snapshot from Subsected. It tweaks the project's settings to make the blocks/path bigger and the subsections narrower, decreasing the density of the design.

The engine uses randomness to subdivide squares, creating a maze-like effect (although this is not a maze-generative algorithm). Calculating and plotting the subsections takes an additional step. I may want to optimize the algorithm itself; it's not my finest (O(2^n) complexity 😬), but people generally won't be working at an `iterations` value high enough for this to be a problem because the output density would ruin the design! Otherwise, it's pretty standard, I mostly spent my time playing with the Blot engine first and then worked on implementing this system :)


leomcelroy:
This is interesting, can you add another dimension of variation to the piece? 

PR #334, opened 2024-03-13
Getting started guide

ozan:
First pass at a new "getting started" guide, which introduces new users to Blot, one set of concepts at a time, by iteratively developing more and more sophisticated drawings of feathers.

@leomcelroy mind taking a look? Still needs edits but thought I'd check in first to make sure this is on the right track ie it's structurally/conceptually sound. It should be accessible locally via http://localhost:3000/editor?guide=start, but also readable enough through the github PR interface.


ozan:
OK @leomcelroy , I made the edits we discussed. Still not 100% happy with it, but at least now it should be more incremental, less overwhelming for beginners, and more consistently copy/pastable with the `tk.` toolkit variable name. I also want through and exported more consistent images. I'd say this is pretty usable as is, and maybe we can iterate it based on beta user feedback.

steps

PR #336, opened 2024-03-16
crazy_star-arthur

AverseABFun:
I made a kind of fractal looking thing that depend entirely on the `size` variable, which as well as changing the size it also can make it look VERY different. I do wish there was dark mode in the IDE, but other then that I like programming for the blot.

Also, if this falls under "simple algorithmic patterns" and won't get accepted, just let me know and I will make something else.


leomcelroy:
This is a bit too simple and probably not drawable. Try to narrow in on a concept for your piece. Start with some hand sketches.

PR #340, opened 2024-03-24
Maze by Neerav

BadPotato1007:
image
I have made a cool maze, that may or may not be possible to solve! it looks really cool ,and I plan to print this on a spcial type op paper that i have, so that the ink can be transfered to a mug! 
image

it should look something like this! would be really cool to get a blot to do this instead of a printer!


leomcelroy:
Could you update this to use the new toolkit? It is described in the editor toolkit on the editor page.

We've been in beta and incorporated the feedback into the version that will go public very soon. Would love your feedback after using both.

BadPotato1007:
uhh will this not work? the newer one looks a lot more complex...

leomcelroy:
This is your piece with the new api.

```js
/*
@title: Cool Maze
@author: Neerav
@snapshot: 1.png
*/

const width = 100;
const height = 100;

const t = new tk.Turtle();

for(var h = 0; h<=100; h++){

  for(var i = 1; i<= 100; i++){
    var yee =  tk.randIntInRange(0,2)
    if(yee > 1){
      t.up()
    }
    else if(yee < 1){
      t.down()
    }
    t.forward(1);
  }
  t.jump([0, h]);
}
t.jump([0,0]);
t.left(90);
for(var h = 0; h<=100; h++){

  for(var i = 1; i<= 100; i++){
    var yee =  tk.randIntInRange(0,2)
    if(yee > 1){
      t.up()
    }
    else if(yee < 1){
      t.down()
    }
    t.forward(1);
  }
  t.jump([h, 0]);
}


tk.translate(
  t.path,
  [width/2, height/2], 
  tk.bounds(t.path).cc
);

drawLines(t.lines());
```

BadPotato1007:
making new pull req with the changed code

PR #349, opened 2024-03-27
blog!-alan

an-Alan:
I decided it would be really cool to be able to make something where the blot bot would draw out words. I ended up making individual functions for every character and had an input string that would feed it in and draw it all out. After taking a long time making the individual letters of the full alphabet I decided to add in some special characters for fun so I can make :3. I had a great time coding even if it was somewhat repetitive, but it brought back memories of messing around with turtle in python. I really hope to see the physical art and see what other changes I can add to make it more fun.

Screenshot 2024-03-26 212706


leomcelroy:
I'm trying to encourage people to be creative with their pieces which I think you could push here some more this is a pretty standard idea that we already have in gallery as well. That being said it's a classic idea to make fonts and there is a lot of creativity to be had there. The font in the gallery is frequently used by other projects so adding another option could be nice. I also like that it's very easy to understand even if it's repetitive.

Congrats if you are a teen you've earned your Blot. You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

We are sorting out logistics so it may take a month or more to have your order fulfilled. This waiting time will come down after the full release.

PR #351, opened 2024-03-28
Fibonacci Spiral-Clay Nicholson

claynicholson:
My artwork, is recreating the fibonacci sequence for blot. You can change the size, number of spirals and if it has boxes. Coding blot was a very fun thing to do with by hack club and I will definitely be doing more designs. 
normal


leomcelroy:
I'm going to close this for housekeeping but please feel free to resubmit in the future.

PR #354, opened 2024-03-30
Create geometricsportscar-Aiden

AidenRheem:
My art is based around a geometric sports car shape, where you can change the different sizes of the wheels that are created using some math radius code to generate the circles. The design of the body is very abstract and the lines are sharp to represent something sleek. There is also a simple window design on the body to make it look more realistic. 
I have never coded in Javascript before, only simple Python scripts.
Screenshot 2024-03-30 122327


leomcelroy:
Thanks for the submission! I think this may be a bit too simple at the moment. How would you make it less abstract?

leomcelroy:
I'm going to close this for house keeping but please resubmit when you are ready. Head over to the blot channel in slack if you need help.

PR #355, opened 2024-04-06
Mountain-EuanRipper

EuanRipper:
I love hiking and so for my art i tried to recreate a mountain scene. Like any good mountain scene there is jagged peaks, sunshine, trees and a bird. The random elements are the shape and number of the random peaks, the position of the bird and trees and the size of the sun. To randomise the shape of the peaks I move the turtle from left to right, with each move to the right the turtle is also moved up or down randomly, with a skew towards the direction of the slope. I also use RNG to determine if a tree should spawn at this position. If the randomness does ask for a tree, I use the drawTree function i wrote and pass in the X,Y and scale arguments. The direction of the slope is inverted at the peak, and the peaks are distributed evenly across the width. I struggled to keep the art in the bounds, and had to think carefully about the limits of the random variables to keep the mountain from overlapping with the sun etc. I think I have successfully prevented this. This was my first experience programming in JavaScript so I spent a fair while getting to grips with the syntax. Once I figured that out I found it very satisfying. I would love to assemble the blot with my hack club.

snapshot3


leomcelroy:
Nice composition. I like how you drew simple things that are easily recognizable as the subject matter. How could you add more depth?

You've earned your Blot. If you are a teen you can [order it here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw). Congrats!

EuanRipper:
Thanks so much! I am a teen and can't wait ;) do you have any idea what the commodity code would be for UK customs tax?  To add more depth i'm thinking of having a river that runs off the mountain and down to a lake in the foreground.
Thanks so much again

PR #358, opened 2024-04-06
Circles

m-anshuman2166:
Not known much to this thing, but playing with circles as my first program was certainly fun !

image


leomcelroy:
Thanks for submitting! I think after some more practice you'll be ready to submit for your Blot. Good work getting started though.

I'm going to close this now for housekeeping.

PR #359, opened 2024-04-06
eclipse-logan

WeasalCrafter:
Hello! In my artwork I have created a depiction of the upcoming North American Eclipse on April 8th! I have some coding experience as I am in high school and have taken an introductory course. The blot API was easy to understand and develop your own intuition in building features. I built a program for the Sprig project before and I can say the improvements to the website are great, the IDE is simple to use and the cloud saving is a wonderful feature. I hope you enjoy my artwork!


leomcelroy:
Thanks for the submission! I think you could add one more artistic element before we merge. How about trying to capture the light that emanates from the edges of the eclipse. How could that be done?

WeasalCrafter:
Thank you for the suggestion! I have committed the changes(I believe), and now there should be sun rays that display based on the percentage of eclipse variable. Let me know if theres anything else I need to do before submission.

BrightTheBackpack:
Hi there! This looks pretty cool.
A couple points of feedback for the rays. For the full coverage ecplise, it would be cool if it looks something like this:
image
Also for the partial coverages, I think it would be cool if as theres less and less coverage, more and more rays appear, instead of having just 3 rays that get larger.
image
image

BrightTheBackpack:
Also, can you complete the header with
/*
@title: 
@author: 
@snapshot: 
*/
That would be greatly appreciated

WeasalCrafter:
Thank you for all your suggestions! I've added them to the best of my ability, the rays now disappear and reappear depending on the coverage, along with a special display for the totality, I made the time change based on the percentage, and finally fixed the header like you asked. Thanks for all the tips.

WeasalCrafter:
> Also, can you complete the header with /* @title: @author: @snapshot: */ That would be greatly appreciated

Sorry if I'm double messaging(I'm not sure how GitHub comments work), but I'm just making sure that I replied to your helpful suggestions correctly and that there's nothing else required of me to complete, thank you!

PR #360, opened 2024-04-06
smile-Ev

DefNotEv:
I made a smiley face that's super customizable. you can edit values for everything and move stuff around. I have a lot of experience coding and I've been to five hackathons + done a lot of projects.

smile1


leomcelroy:
I think this may be a bit too simple at the moment. What are some ways to make a face like this more beautiful and intricate?

leomcelroy:

PR #364, opened 2024-04-07
`flower-coda`

Codaea:
image (2)

looking through active pull requests, flowers seem a little common, but maybe this is a new take. 
for this piece, I was interested by existing take on geometric flowers that I found online, that where drawn on a pen plotter, along with the cool trees people managed to draw with them. the part of simulating something organic using a algorithm is really interesting, and something I definitely plan to work on more.

at the beginning, it was definitely a challenge to figure out how to keep the stems of the flowers from being to close to each other or too unnatural, (like sideways) and I ended up putting limits on how far each point could go. I also took inspiration from the default example with the flower themselves, and when playing around, I noticed certain degrees like 180 and 300 make it a triangle or pentagon spiral. I also spent a good bit of time making a point marker that makes "X"s for debugging. definitely planning on expanding this in the future, does blot robot let me use multiple colors if I go in layers?


leomcelroy:
Thanks for your submission. We're trying to level up the level of flowers people share. How might we approach that here. I think you should start by finding a real flower to take inspiration from (or an image of a real flower). Take a good look at it and tell me what you see.

Codaea:
I kind of see orchids, but without the leaves.

leomcelroy:
I mean find a model image to work from.

leomcelroy:
I'm going to close this for housekeeping. Feel free to submit again when you are ready.

PR #366, opened 2024-04-09
SnowinOnASnowman-AidenP

PilotOfGolf:
This experience in coding has proven to be very educational. I learned a lot from this, and hopefully, this knowledge will stay with me. There were some hiccups along the way, such as the documentation being incorrect, however thanks to the super-friendly people on the HackClub Slack, I was able to mitigate the issue, and the documentation was subsequently fixed. 

I wanted to take a math-heavy approach to this, instead of just using a simple turtle. I think as a result my artwork suffers from not looking amazing, but I learned more this way. 

I spent around 12 hours doing this in total, and I am proud of this, so I hope this gets the seal of approval. 
Thanks!
snapshotGallery


leomcelroy:
Thanks for the submission!

I like that it's a composition, and the code is nicely structured. I think you've earned your Blot. If you are a teen you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

As for leveling up the snowman, do you think you could take any inspiration from this?

Screenshot 2024-04-09 at 3 03 51 PM

PR #367, opened 2024-04-09
nothome-Francielly

frananisio:
The coding experience was a journey of learning and discovery. Through the process of experimenting, tweaking, and optimizing, I was able to hone my programming skills while creating a digital work of art.

This project is a testament to the power of programming - a tool that not only solves complex problems but also serves as a means of artistic expression. Through code, I was able to bring my vision to life, creating a serene landscape of a house next to a tree.

With this drawing and its title, I leave room for interpretation about what I would like to express with the drawing.

notmyhome


leomcelroy:
Thanks for the submission. I think this may be a bit too simple at the moment. Could you add some perspective to the house or make the tree look more realistic?

PR #370, opened 2024-04-09
randomcatgenerator-thebaconpug

TheBaconPug:
I have created a very customizable cat creator. You are able to modify every aspect of the cat using the constants at the top of the file. The code isn't amazing but it gets the job done. 

This project was fun and I had a good time while creating it.

cat1


BrightTheBackpack:
Thanks for your submission! This looks awesome! Do you think you could make it so the pupils never leave the eyeball? I think that would make the piece better.
image

leomcelroy:
I like the playful haphazard-ness. Do you think you could add a body?

leomcelroy:

PR #371, opened 2024-04-09
SimplePc-Propiggy009

Propiggy009:
This artwork is a simple picture of a PC with a monitor, keyboard, mouse and speakers. I dont have that much experiance in coding with JavaScript but I do know how to code in Python Turtle. This is why I choose to do this. Below is a picture of the artwork.

Screenshot 2024-04-08 214914


BrightTheBackpack:
Thanks for your submission! This looks really cool. The drawing doesn't seem to be parametric. There should be some parameters to adjust or the ability to run it multiple times and get new flowers. Also, I think it would be cool if you added some more stuff to the pc tower, like a power button on top, or fans on the side.

Also make sure none of your snapshots have spaces in there names.

leomcelroy:
This is really great. The only thing you need to do to make it Blot worthy is make it parametric. That means there should be some parameters at the top which can be adjusted to change aspects of the image.

Propiggy009:
I have changed my artwork to have more additions like new different things to displays, computer fans and a power button.  This is just one of the 2 other things that the screen displays. Below is the picture. I have changed the files accordingly.

Snapshot3

leomcelroy:
Nice improvement!

If you're a teen, you've earned your Blot. You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw). Congrats!

Propiggy009:
Thanks for the news!

PR #372, opened 2024-04-10
puzzleGenerator-maxElder

Hoggie5:
I made a small piece of code that generates a random puzzle from the given inputs. I have a fair bit of experience coding, including making games with sprig.
screenshot3


Hoggie5:
> Nice! I think this is Blot worthy. You just need to make sure the file has the right metadata.
> 
> How would you extend this to make the puzzle pieces more irregular?

I could probably change the connector node piece things to be different shapes if this isn't enough. I'll double check the metadata to make sure nothing is wrong.

leomcelroy:
It could be developed more but I think you've earned your Blot. Congrats you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #374, opened 2024-04-10
cirangle-Lovidu Add files via upload

Raptor-Coder:
My artwork is a combo of circles and triagnles randomly places to form some nice asthetic images. I don't have too much experience coding and am currently still learning. I found this project to be a fun way to explore java script (this was my first javascript project so I had some help from a friend) and create something cool.
snapshot1


leomcelroy:
Thanks for the submission. I think this is lacking intent. Try to think of a composition you would like to make in advance. I recommend targeting something non-abstract.

PR #383, opened 2024-04-10
Shamrock_generator-Eoin

covertimpala:
The code generates a unique shamrock with several adjustable parameters.
I have experience programming with python, however very little experience with js.
It took a while to get the parameters right so the shamrock pattern is generated but it was well worth it.
image1


covertimpala:
Added shadow generation and a water ripple effect below the shamrock in order to increase the complexity of the scene.
Preview

leomcelroy:
That was not the direction I thought you would take it. Nice work! It looks really cool.

Congrats if you're a teen you've earned your Blot. You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #385, opened 2024-04-11
Cursorwall-smashmaster

javaarchive:
Been a while since i've worked with any turtles (last time was when python 3.4 was the latest python iirc), but this was fun to see again, and unlike python turtles, the drawing is instant. 
image
Originally I had some more graphically intense ideas, but I wanted to actually make something "drawable" in a reasonable amount of time. I ended up making this accidentally and thought it looked aesthetically pleasing and it reminds me of some calculus slope fields except it's not just boring lines. Each one of these "cursors" is randomized on run but they're roughly controlled in direction for now (unless you edit the angle randomization) for the aesthetic effect of "turning".


leomcelroy:
I like this but the code involved is a bit too simple. While not always a great metric what would you add if it had to be 150 lines long?

javaarchive:
thinking of updating code to do some math calculations to have things face a random point, will update PR in a bit =)

leomcelroy:
I'm going to close this for housekeeping. Feel free to resubmit in the future!

PR #388, opened 2024-04-11
Cheese-UserJ

UserJHansen:
This wonderful little project makes delightful little (non-edible) cheeses.

I really wanted to focus on getting the perspective right, which is why the camera control got in there. I've always been inspired by 3d rendering toy projects, and while this isn't even close to that kind of stuff, thinking in 3d still twisted my brain in a way that forced me to step back and try again quite a few times.


leomcelroy:
Nice work! It's recognizable as cheese and you used some great techniques to get the 3D effect.

Congrats, if you are a teen you've earned your Blot. You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

Some feedback to keep improving this. The holes are distinctly 2D, how could we give them depth? And when the cheese wheel occludes itself the perspective breaks (large cheese angles).

PR #389, opened 2024-04-11
Clean up IDIOMS

UserJHansen:
Idioms seems to be quite outdated, this pull request might even need more stuff added to it, but it at least removes the references to the old toolkit.


leomcelroy:
Thanks for contributing to this. I think the PR is a bit outdated now so I'm going to close it for clean up.

UserJHansen:
@leomcelroy, it doesn't look like any changes have been made to fix the code in the IDIOMS file, is it possible to merge instead?

PR #390, opened 2024-04-11
MandelBlot-Andrew

GoCommitOwO:
My blot program generates the Mandelbrot set at varying degrees of recursion which is definable by the user. I have been programming on and off for something like the last 8 years, but I didn't really fully get into it until some time in the last ~2 (before that I was stuck doing stuff like Scratch and I didn't really feel much motivation to program!)

I have some concerns as to how easy this would be for Blot to draw, but there are several ways I could simplify this if it ends up being too complex with how I did it - or, more likely, this could just be run at a much lower level of detail and then it would certainly be able to draw it. On a related note, if you set MAX_ITER to any value below like 10, it starts to get really blobby and odd with how it looks - I think that's pretty cool!!

I also would like to figure out how to do some proper weighting for the random selection, since it's obviously way more likely to give you a super well defined set if it's just rolling a number from 5-80 iterations, but that's not something I'm very confident in coding in a Javascript-like language - I'm more of a Python user, so even this was a bit of a stretch for me!

100-Recursion
^ ^ ^ This is what extremely fine recursion looks like

15-Recursion
^ ^ ^ This is only 15 iterations, which is where it starts to get into the wonkier territory

5-Recursion
^ ^ ^ This is what it looks like with extremely few (5) iterations. Pretty neat!!


leomcelroy:
Thanks for the submission! A Mandelbrot set isn't the most creative as it's a pretty standard math visualization but also because it's so common it will be nice to have in the gallery. You have a good implementation and there is some creative interpretation in how you approach rastering.

I think you've earned your Blot! Congrats, if you are a teen you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

If you want to contribute more a guide to iterated function systems would be really useful!

PR #394, opened 2024-04-11
grassOrHair-mint

mintexists:
This artwork is inspired by this artwork https://inconvergent.net/generative/linetrace/ but is my own interpretation on the idea. The first line on the left is made of small steps forwards followed by randomized turns. The next line follows the same path as the first, but adds its own random turn on top of the previous one. This randomization allows the lines to change dramatically, or for the randomization to cancel out forming straighter lines. Another element I randomized was the size of the randomization of the turn after each step, which with smaller values can create smoother, more flow-y works, and with larger values can create chaotic tangles and knots. One aspect of this that I enjoyed was that with smaller random values, the results looked almost like a field of grass or hay blowing in the wind, and with the larger values it looked like when my hair gets tangled in my hairbrush and forms a ball. 

I really enjoyed the experience making this! I enjoy seeing generative art, but have not had much of a chance to make my own. It was very interesting imagining the paths that the Blot would take when drawing the lines. One struggle I encountered was that I could not find the documentation at first, because the blue colors in the Help area were too similar and I thought it was all one section which said "Workshop Toolkit", rather than two sections. I did struggle with the implementation of my idea, and had several bugs where the turtle would go in the wrong direction, but I took a break and was able to figure it out afterwards. 

SCR-20240411-oang


leomcelroy:
I think this is a great start but it feels a bit like a technique demonstrations without fully incorporate it into a complete piece. 

Do you think you could lean into the grassy field? Maybe something like the drawing below.

Screenshot 2024-04-12 at 6 47 11 AM

leomcelroy:

PR #395, opened 2024-04-12
transformationTelevision-Eric

EricKugel:
I've always found the moire effect weird. I've also always found old TVs weird. This piece shows the moire effect on an old TV screen by using a non-linear transformation. Weird! When I had the idea and looked through the docs, I was surprised to see that you only had linear transformations available. I'm doing an independent study in linear algebra right now though, so several hours and probably a hundred tabs later, I finally understand how to implement it myself. If you want me to make a pull request to add a non-linear transformation function to the source code, let me know! I'd be happy to.

My main goal was the moire effect though, and I'm not sure if it'll show up on paper. I'd be interested to see how it looks once I draw it, and whether I'll have to overlay it with a parallel grid or not.

I've been coding for quite a while now, but my main languages are more of Java and Python. I found the map function in JS super cool! I love when I get a chance to use it.

Thank you for doing this -- I love what this repo is doing -- and thanks for your consideration!


EricKugel:
Just realized my transformations might not be non-linear... the term I was looking for was projective transformations.

UserJHansen:
I'm currently doing some janky transformations on my cheese project for blot, how easy do you think it would be to use these transformations in another project, and could I use it as reference?

EricKugel:
> I'm currently doing some janky transformations on my cheese project for blot, how easy do you think it would be to use these transformations in another project, and could I use it as reference?

I saw your cheese project lol! It's pretty cool!

It would be pretty simple to replicate my code, I think. The way I do it just transforms the entire drawing, which I'm not sure you want, but you could isolate certain polylines to transform too. Yes absolutely feel free to use it as a reference! Let me know if anything doesn't make sense.

UserJHansen:
Awesome, thankyou! I hope this fixes the holes vacating at high camera angles 😭

leomcelroy:
This is a great submission. It's a very powerful technique and I'm glad to have such a nice demonstration in the gallery. You also used it to nice effect to actually make a recognizable and nice looking image. Good work!

If you're a teen you've definitely earned your Blot! You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw). 

It'd be amazing if you wrote a guide to using projective matrices for drawings.

EricKugel:
Thank you so much! I will start working on a guide, that will be fun!

PR #402, opened 2024-04-13
Planet and Moon

blobbybilb:


leomcelroy:
Thanks for the submission. I like that you included some details on the planet and moon. I feel the distribution and star shape could improve some more. It also seems to take abnormally long to generate. 

I think you've earned your Blot though. Congrats, you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #403, opened 2024-04-13
LuffyArtist-Nathan

agentbobjelly9228:
This artwork is a drawing of Luffy eating a burger. The brightness/contrast of the drawing will change each time you run the program. I had a great time coding this project. I had to incorporate some python imaging techniques before transferring to the JS. 
luffy1


Jobin-S:
How do you generate the data points?

agentbobjelly9228:
I used python to convert the data from an AI image before transferring it
to the blot editor.

leomcelroy:
Thanks for the submission. Your Blot program should generate the image not just render an image from somewhere else. We do have some rasterizers but the aim is for them to be transformative. I think the simplest versions (like pixel present or not) are already in the gallery.

PR #405, opened 2024-04-13
Satellite-James_Ma

JamesTMa:
This artwork is a simulation of a satellite moving through a system of a star and a more massive black hole. Although it is entirely mathematical (as in it purely follows gravitational laws), I find that the patterns one can make with it are beautiful in their own right.

I am a current Junior in high school. I have been coding in one way or another since elementary school, but I only started taking classes for it within the last two or three years. Initially, I learned most of it from my brother, and I am now capable of writing in Swift, Python, Java, C++ (in descending order of proficiency). Javascript, however, is entirely new to me. I've always enjoyed independent projects, like Wordle and Tetris in Pygame.

Hopefully this format works:

Not_this_one


leomcelroy:
Thanks for the submission! I like the idea but maybe you could use it to make something a bit more aesthetic.

Maybe you could take inspiration from something like this depiction of spatial distributions of singularities in a Lagrangian.

 
Screenshot 2024-04-22 at 1 37 17 PM

(from https://www.researchgate.net/figure/Spatial-distribution-of-singularities-in-the-Lagrangian-and-Eulerian-Cosmic-Web-The_fig5_315682334)

leomcelroy:
I'm going to close this for housekeeping but feel free to resubmit later.

PR #407, opened 2024-04-14
LiveLikePicasso-SamratBose

samratbose10:
Art always fascinated me a lot specifically the sketching, I never been so good at it. So, when I first heard about the blot I thought I could be like Picasso(one of the GOAT)😎. 
But here comes the hard part, I'm never tried js, although I made some arduino projects, so I knew I can build it. It took me a week to learn js and made my first project "ThePlanets" and then tried to build a code to do sketching, but it was so much tough for me to do it alone so I sought the help of a friend, who helped me to build this project, it took us more than 15 days to build it.

So my project is "LiveLikePicasso" through which anyone could make sketches(like Picasso 😉). This project convert the black and white img to a blot code and can be used to make sketches.

snap
for more examples you can check out the snapshots folder.

P.S- It's like a dream come true for me.
hewillalsobehappy


kcoderhtml:
This is super cool!

leomcelroy:
You need to do the image processing in JavaScript. I'm probably going to add some image uploading this week but an approach others have used is adding an HTML file to the submission that will allow you to drop in an image and convert it to a format that can be pasted into your Blot program. 

I think the raster approach may not be transformative enough either. Think about ways you can really modify the image and not just render it.

I'm going to close for housekeeping but feel free to resubmit when you are ready.

PR #408, opened 2024-04-14
DesertSunset-Augustin

piggy200:
My artwork is a desert sunset with perspective. In art class some time ago, we learned about perspectives and how things get smaller as they go away, and I wanted to draw that with blot. I am a beginner coder, and I believe that blot is really nice because it allows you to directly have hands-on contact with code. For example, I found a clever way to use "for" loops to create the lines in the road. Also, I found that random variables and iterating points were fun to use. To make the mountain jagged, I used the iterate points function (as shown in the leaf tutorial). I think blot is really cool and it has a lot of potential. I really want to learn how to randomly generate fractals like the sierpinski's triangle, etc.

desert3


Jobin-S:
awesome dude

PR #416, opened 2024-04-16
generativeMountainArt-hephaestusHex

hephaestushex:
This project generates perspective mountain art. I've always liked drawing perspective art with roads, and with mountains close by to me, I made this art generator. Almost everything is user inputed in this art, except for the mountains, representative of the fact that mountains are works of nature, and humans cannot control it. My experience with coding is limited, as I only know to program in Python , C#, and C++, and my experience with JS is minimal. I enjoyed reading the blot docs to make this art work.
snapshot_00


leomcelroy:
I like these perspective mountain scenes. Nice work using clipping.

Congrats if you're a teen you've earned your Blot. You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #417, opened 2024-04-16
pizza-Kai

kaiyuran:
My artwork is a drawing of a pizza. It has a crust, pepperoni and mushrooms. They can be adjusted in size to change the image and the toppings are randomly generated on the pizza. I have experience coding in the past but have not had much experience working with graphics. 
snapshot1


leomcelroy:
Thanks for the submission. I'm wondering if we could add more detail to the pizza. When looking at line drawings I notice a lot add details to the cheese. Here is an example:

Screenshot 2024-04-22 at 1 54 11 PM

I think the dripping isn't the most appetizing but it's an interesting effect. How could you achieve that?

kaiyuran:
I have updated my art to include some dripping cheese around the pizza using Catmull-Rom splines. They scale with the Pizza and I made the sizes of toppings/crust a ratio to be scalable as well.

leomcelroy:
It's still quite messy looking. Not so much in a cheesy way but more so like the drawing is haphazard. Try using cut and cover to add occlusion to the piece.

leomcelroy:
I'm going to close this for housekeeping. Feel free to resubmit in the future.

PR #418, opened 2024-04-16
modify readme to inverse background-less images on dark mode

hyusap:
two of the images are hard to view on dark theme, see below
image
image


leomcelroy:
Thank you!

PR #421, opened 2024-04-17
raymarcher-lolgeny

lolgeny:
I wrote a raymarcher, using small crosses to create different pen thicknesses, an SDF made of mostly spheres and a link object (with smooth blending), and ambient + diffuse lighting.

Parameters include the position of the camera, the lights, the level of detail, and position of spheres. Additionally, random spheres are generated in the background for effect.

It was very educational to make as I wanted to learn more about raymarching and SDFs.

img2


Babyyoda777:
Very cool 👌

lolgeny (commented):
Removed commented out code

PR #422, opened 2024-04-17
racket-ahowden27

ahowden27:
My program creates an important item from one of my favorite sports - a tennis racket! In the first few lines of code, there are variables that can be changed, including the amount of strings on the racket, the size of the racket, and the angle that the racket is displayed.

The coding experience was really fun, althought it took quite a few iterations before I figured out how to get the strings to not extend past the edge of the racket. It's always a great experience learning new syntax, and I definitely want to see what else I can make with the machine. Thank you for giving us this opportunity, it definitely gave me the break from schoolwork that I needed!
racket
racket2
racket3


leomcelroy:
Thanks for the submission. Do you think you could add some thickness to the strings in the racket and make it look like they are woven together?

ahowden27:
Definitely, I will work on that!

ahowden27:
I made the revisions, but can't figure out how to commit the new screenshots so I just dropped them in the description... let me know if I should make a new PR if you need the images that way.

PR #426, opened 2024-04-18
DesertSunset-Augustin

piggy200:
My artwork is a desert sunset with perspective. In art class some time ago, we learned about perspectives and how things get smaller as they go away, and I wanted to draw that with blot. I am a beginner coder, and I believe that blot is really nice because it allows you to directly have hands-on contact with code. For example, I found a clever way to use "for" loops to create the lines in the road. Also, I found that random variables and iterating points were fun to use. To make the mountain jagged, I used the iterate points function (as shown in the leaf tutorial). I think blot is really cool and it has a lot of potential. I really want to learn how to randomly generate fractals like the sierpinski's triangle, etc.

desert2


leomcelroy:
I love it! Nice use of color in the render too even though it won't come up in the drawing (unless you color it in). Maybe you could explore switching pens and adding colors with your Blot.

One piece of feedback is given the scale of the mountains the road line should become very small and have quite a few dashes in the distance. Otherwise I would approve now but your file structure is pretty messed up. Could you clean up the submission folder and I'll merge.

leomcelroy:

PR #427, opened 2024-04-18
Ear-Cindy

playfun24:
My artwork is based on the figurative image of an ear. During the coding process, I adjusted the coordinates of the points so that it can accurately display the shape. I added variables and randomness that translated and scaled the original image to produce multiple variations.


leomcelroy:

PR #430, opened 2024-04-19
SineArt-Ruben

pricken45:
This is my Sinusoidal wave image generator. To make your own image, generate it with the imagegenerator.html file which gives you a 2D array that you can paste in the index.js file. The code generates really cool looking (according to me) results with quite a bit of ability to customize the result by changing resolution, step size, frequency multiplier, amplitude multiplier and, of course, the image itself. This was a fun project with creative coding and I learned quite a bit from it, especially the data sine wave modification part.

The image array generator can be found [here](https://imagearraygenerator.netlify.app/)
bird


leomcelroy:
These image transformations are cool but submissions have to be runnable by other people too. One thing you could do is to add an html file to the folder that allows you to upload an image and converts it to a format that Blot will accept. I may add this to the editor itself actually. It would work like dropping in an SVG now but instead of returning polylines you would get a bytearray.

pricken45:
Hi. The API used is public for everyone to use. It should be bo problem for other people to use it since it is hosted publicalöy 24/7

leomcelroy:
Nice I think I already sent you the Blot form for your second submission. Thanks for sharing more work.

PR #433, opened 2024-04-22
blotMaze-Sophia

evanss2025:
This is my Blot project, blotMaze! I used Blot to make a maze with the goal to get to the crown. This was my first time using javascript for art so I was super lost at first. But, with time I finally got the hang of it, so although it may not be the most detailed thing ever, I am super proud that I challenged myself to learn something new.


leomcelroy:
Do you think you could challenge yourself to generate a new maze on each run?

leomcelroy:

PR #434, opened 2024-04-23
LiveClock-Ruben

pricken45:
This is my real time generative clock that gives you the current time where you're located. For fetching the time it makes use of javascript's built in Date class and passes along the angles to the turtle used for the drawing.  To make all the circular items I made use of polar coordinates that are converted to cartesian coordinates for the actual drawing. I find polar coordinates to be really interesting and quite fun to work with. The generated clock is customizable in size, stick thing lengths and position and if you want the seconds hand of the clock. This was a fun little project!

image


leomcelroy:
Thanks for the submission. I feel this piece could be more artistic but I'm going to accept it because it's a nice demo of using browser APIs and some basic math to make something coherent.

Congrats you've earned you Blot. You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #436, opened 2024-04-24
SineSpiral-Ruben

pricken45:
This is a project similair to the "SineArt" project I made a few days ago. This version however makes one continuous spiral difformed by a sine wave with variable frequency and amplitude. This project makes us of polar coordinates and radianses and those are quite new to me. I learned even more from this than the "LiveClock" project I made yesterday. I am the most proud of this one since the art is more aesthetically interresting with more variation avaliable. I also did some fiddling to make the scale variable to work properly!

adamsandler


leomcelroy:
I would still prefer if you added the website into the PR in a single html file. That being said nice work on the piece. It's transformative enough to become it's own thing and not just a renderer.

PR #438, opened 2024-04-24
racket-ahowden27

ahowden27:
This is a new PR to accomodate the suggestions:

My program creates an important item from one of my favorite sports - a tennis racket! In the first few lines of code, there are variables that can be changed, including the amount of strings on the racket, the size of the racket, and the angle that the racket is displayed.

The coding experience was really fun, althought it took quite a few iterations before I figured out how to get the strings to not extend past the edge of the racket. It's always a great experience learning new syntax, and I definitely want to see what else I can make with the machine. Thank you for giving us this opportunity, it definitely gave me the break from schoolwork that I needed!
racket


leomcelroy:
Nice! When I mentioned weaving I was thinking add some thickness to the strings and show them overlapping each other. This is a different approach that seems to work well too though.

Congrats, you've earned your Blot. You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #440, opened 2024-04-25
Mario'sMysteryBox-DevonShunn

CondolentFire10:
I have no idea how to use GitHub to be completely transparent-  I think all my responses to questions should be attached to the fork itself?  Let me know please :)


leomcelroy:
Thanks for the submission. I think you may want to revisit this concept. Right now you're using Blot to render a preexisting image. Try to write a program that generates an image. Incorporate randomness so you can get a new image each time it is run.

leomcelroy:

Issue #392, opened 2024-04-11
bt.cover mangles polylines1 when polylines2 fits entirely within its boundaries.


Elijah-Bodden:
See [here](https://blot.hackclub.com/editor?shareId=3f1c6be6-d600-42ce-9262-a233f70db65f) for an example. This is a more general problem - whenever the outline of the covering shape fits completely within the covered shape, vertices on the covered shape get set to apparently random positions off-canvas. If the coverer is outside the shape, or it overlaps with its lines even a little bit, no problem. But as soon as they're over each other, this bug happens.

Elijah-Bodden:
Looks like I misunderstood the purpose of cover, subtract does what I was thinking cover should.

Elijah-Bodden:
I traced the bug to the `addToNdp` function in cutCover.js. Replacing calls to it with ndp.push seems to fix the problem but apparently we need `addToNdp` for other reasons

PR #441, opened 2024-04-25
Replace addToNdp with ndp.push

Elijah-Bodden:
Since addToNdp causes issue #392 and replacing it with vanilla ndp.push doesn't (seem to) break anything, provisionally replaced it.


leomcelroy:
Thanks for looking into it. Let's see if any problems arise.

PR #443, opened 2024-04-26
City @ Night - Winter Edition

alexdevz:
This is a pretty straightforward/simple thing, but I decided to generate a city at night, in the winter with snow falling. Each generation has a different set of buildings and snow placement. :)


leomcelroy:
Thanks for the submission! Could you make the windows stay inside the buildings?

alexdevz:
ill try :)

PR #445, opened 2024-04-26
Pokéball - Elia

eliasangelss:
Over the past two weeks, I've been working on coding a Pokéball using the Blot editor. It was styled to look as similar as possible to the original art, with a two-layered button and the slightly off-centering that Nintendo does to match :') To suit the criteria, I added a random number generating function that customizes the design of the Pokéball depending on the number it lands on. Each design is a different version of the ball that's available in the games (particularly some of the balls that were first introduced in Red, Blue, and Yellow.) A standard Pokéball is generated when 1 is selected, a Great Ball is generated when 2 is selected, and when 3 is selected (else) an Ultra Ball is generated.

Before working on this project, I'd never coded anything using JavaScript before. I had to learn the basics of JS using the toolkit explanations and some freeCodeCamp articles, and I heavily underestimated how difficult it would be. At one point, I didn't think that I'd ever finish it in a way I was satisfied with. But with the help of the community (and some caffeine...) I managed to create a piece of art that I'm really proud of. I'm looking forward to being able to teach others how to use Blot at Evergreen next weekend! 

image


leomcelroy:
Very nice! I noticed your persistence in the #blot channel. I think it paid off well. I think these would look great colored.

Congrats! You've earned your Blot. You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #446, opened 2024-04-27
3DRoom-AgrimBansal

Agrim-Bansal:
In this program, I have coded a 3D study-room with cuboidal objects like books, shelves etc. This 3D room is then rendered to a 2D picture algorithmically with respect to a randomly varying point of view. 
Having no experience with art or modelling at all, I am quite proud of it.

Screenshot from 2024-04-27 17-18-18


leomcelroy:
Really nice, I like it too!

Could be interesting to add some hidden line culling. But it's already great.

Congrats! You've earned your Blot. You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #447, opened 2024-04-28
City @ Night - Winter Edition

alexdevz:
This is a pretty straightforward/simple thing, but I decided to generate a city at night, in the winter with snow falling. Each generation has a different set of buildings and snow placement. :)

The code is easily configurable for anyone to adjust to their likings!

image


leomcelroy:
Thanks for updating it! I think the snow works better as stars (and consequently being behind the buildings). It's a simple piece but recognizable. I think it could be developed much more by using more than just basic shapes (rectangles).

That being said, it's nice to have some basic examples in the gallery and for a beginner the composition is approachable. 

The snapshot name is messed up. Can you correct that?

leomcelroy:
There still seems to be an issue with the img title in the metadata.

leomcelroy:

PR #448, opened 2024-04-28
Character - Naman S

Nam2302:
This artwork is a figurative piece of a character, as each person has their unique set of attributes and features. I had an amazing time coding this piece and learning how to use Blot through the provided toolkit. I learned to use my previous knowledge in Python to apply into the program, while using new things such as functions and constants.  

Image:


leomcelroy:
This is simple but it's nice. I'm noticing though that you're getting this fill effect from a bug.

You can see it here.

```js
drawPerson();
drawLines([ ...finalLines, ...body.path, ...arms.path ]);
``` 

`drawLines` takes a list of polylines first, you passed multiple polylines which is being interpreted by the function as the fill arguments.

When you draw lines properly it looks like this.

Screenshot 2024-05-02 at 1 22 53 PM

It is possible to use the undocumented fill arguments to draw the shape as intended. Can you make it work as a line drawing though?

leomcelroy:

PR #451, opened 2024-04-29
ChargedParticles-Jackson

jackson-smith-a:
image2
image1

Hello!

I saw your project on Hacker News a while back and decided to give it a go. I don't know anything about algorithmic art OR programming in javascript, so it was an interesting challenge! I struggled to think of something interesting to draw, so I decided to think about what I _do_ know a little bit about: physics!

I recently had the opportunity to take a physics class going over the basics of Maxwell's equations at a local community college. I was really struck by the beauty of Gauss's law, in particular, describing how the flux of electric field through any surface is directly proportional to total enclosed charge, regardless of how it's positioned. From this law follows Coulomb's law, describing the force between two charged particles.

My program takes a configuration of static charged particles and a single unbound charged particle, and draws its motion through the electric field the static particles generate.

The hardest challenge for me as figuring out a good way to draw the motion of the particle itself -- I briefly considered having a dot at intervals spaced evenly in simulated time, but I decided a single line would be prettier. I initially tried to record every 10th data point and interpolate a curve through them, which was in hindsight not a great idea. It was surprisingly slow at higher resolutions, and was really just throwing away data. I eventually came to my senses and recorded every data point into a polyline and displayed them all together, which worked quite well!

Even if I don't get accepted, I'm quite proud of my little program, and was really impressed at the pretty patterns it can make :)

Thanks for this awesome project! I doubt I would have ever explored this without it.


leomcelroy:
Great visualization! Thanks for the context in your PR.

PR #455, opened 2024-05-01
aLittleFlareIntoYourExpression-Nirvaan

Owlfx:
Hi, my name is Nirvaan, and coding started as a hobby but turned into something more than that. I am fluent in Python, and learning C#. I have created many cool things over my 1-year journey. I've attached those for you. My artwork reminds me that while you might go through ups and downs as the emoji changes, we are still in the same program, and can calm our life down. Sorry, mine is not the best. Tried hard on it since I do not know much about js and turtle :).

.
snapshot1
/github.com/hackclub/blot/assets/136026268/3d0739d6-3532-49b5-b8ff-43bea3987a2a)

World
Amazin


leomcelroy:
Thanks for the submission! I think this may be a bit too simple at the moment (and similar to some other pieces we have).

How could you make the emoji more unique? Did you have any other concepts you were exploring?

leomcelroy:
I think the gallery may be getting a bit crowded with emojis so the bar is going up for what you'll have to do if the piece is similar to existing pieces. I would recommend exploring a new concept or adding an element which is non-obvious how to draw algorithmically, like hair.

PR #459, opened 2024-05-09
SoftFractals-Ram

Screen-sLaYeR:
Nice Journey for this Fractal Generation
Learned Something New
soft_fractal1


leomcelroy:
This is an interesting piece. The code seems to rely on a lot of AI assistance but to my eye it seems a bit more complex than what it could spit out completely on it's own. I like that it's not immediately obvious what the pattern is.

I think you've earned your Blot. Congrats! [You can order it here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #461, opened 2024-05-11
The garden!

HarjyotS:
This artwork is a beautiful scenery of a garden that randomly generates different flowers and grass on each run I had a great time programming this and learned a lot about the blot ecosystem. I have used my prior experience with Python and HTML to understand the functions and how to use them.


leomcelroy:
I think this may not quite be there yet. Could you try to make the flowers with outline strokes? The code also seems to rely on AI assistance quite a lot. The idea of drawing a field is nice.

PR #462, opened 2024-05-11
raytracer-hypertntclown

HyperTNTClown:
I wanted to finally learn something new in the field of graphics programming after having mostly learnt webdev and backend stuff, so I wrote a raytracer. It is able to render spheres & triangles, so essentially everything you'd need to construct anything. It renders in two modes: one with an ascii-art-esque post-processing filter, and the second one is just the basic raytraced output. You can switch between those using a boolean switch placed conveniently near the top of the file.
There are also a number of other options to play around with, like for example the resolution, the recursion depth for reflections, line width & letter scale for the ascii-art, as well as options to change the objects in the rendered scene.
Learning the basic concepts of how a raytracing works was quite interesting as it always seemed sort of like a magical black box to me.
I did already have some moderate experience in javascript which definitely helped in the creation of this, but I'm fairly new to graphics programming.

img
img1


leomcelroy:
Another classic graphics program. Nice work with it! I challenge you to make an artistic composition with it now.

You've earned your Blot though. I'm impressed with how you made use of the fill commands. [Here's the order form](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

Lightshayan:
how'd you make color?

BrightTheBackpack:
drawLines() supports color! check out https://blot.hackclub.com/docs

PR #463, opened 2024-05-12
Algorithmic Square Randomizer

shaurio:
Hi, my name is Shaurya. I am 14 years old and in 8th grade, this is my submission. My dad suggested this to me, so I gave it a try. I've dabbled in lot's of different programming languages and have been using scratch since I was 8-9. This looked very interesting so I tried to come up with something that would change each time it ran. However, my project does not meet the criteria, as I intended it such. Before pouring my time and effort, I would like to know any suggestion, or rather feedback you have for me to improve/enhance my project to where it is deemed as fulfilling the criteria. I have some idea for the future of my project, I have listed them in the project file itself. Please take a look, write me back some feedback and or suggestions. This way I can know for certain whether my project is worth pursing, or if I should drop it and start another project with a differing objective. 

square_randomizer_main
square_randomizer_1
square_randomizer_2
square_randomizer_3


leomcelroy:
I think this is a great start. Maybe you could add more projections of polygons that work the same way but don't overlap on the front face. Then fill the image this way. Does that make sense?

shaurio:
I’ve had a week full of sports and activities, and got around to finishing this over the weekend. I spent a couple hours and took it from one square, to as many as someone would want. I experimented with having 3 squares initially in random positions, then converted it into a scalable function. This allowed me to make more projections, and fill the screen without making it cluttered. I then wrapped it in a loop so that I can call upon the function as many times as wanted (i.e., by tweaking the cpx range) to create a cool effect.  Here are some screenshots of the latest runs, tell me what you think

square_randomizer_main_v2
square_randomizer_4
square_randomizer_5
square_randomizer_6
square_randomizer_7
square_randomizer_8
square_randomizer_9
square_randomizer_10
square_randomizer_11
square_randomizer_12
square_randomizer_13
square_randomizer_14

leomcelroy:
Can you make it so the shapes don't overlap with each other. Trying using the cut and cover functions.

leomcelroy:
I'm going to close this for housekeeping but feel free to resubmit when you are ready in the future.

PR #464, opened 2024-05-13
MountainSunset--AlbertM

AlbertMundadan:
A jagged mountain/cliff with a bright red sun in the background, birds flying in the sky, and a tree in the foreground. Inspired by Chinese ink wash paintings. The jagged edges of the cliffs are random and the number of birds in the sky can be modified. The style of the tree can also be modified by changing the angle and number of iterations. 

I have a moderate amount of experience coding but mainly in python. This is my second time using JavaScript.

Mountain1
.


leomcelroy:
You took some classic computer drawings and composed them nicely. I think the color actually does a lot here. It feels like it needs something else to connect the mountains. Maybe a river or just more mountains.

It earns a Blot though. Congrats! You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #465, opened 2024-05-13
rainingcloud-blasty

BlastyTheDev:
My artwork is a cloud which is shown to be raining. The shape of the cloud consists of curves, like clouds you would probably draw if you were to draw clouds. The program uses multiple parameters you can change to affect the cloud's shape, size and the amount of rainfall there is.

Somehow this took quite some time to make, around 3 or 4 hours I think...

My experience in coding is quite specific to certain libraries I use, which started with using Java to make plugins for a Minecraft servers between friends (Paper API), then I started learning web development and I understand the basics of HTML, JS and CSS. I also have some experience in Python, which started a bit before my Java experience, but I never used or practised using Python until a coding class I had in middle school. I would say I am "intermediate" level in programming.
image


leomcelroy:
I really like the idea of a rainy cloud. I think we could make it look a bit less flat though.

What can you learn from the drawing below?

Screenshot 2024-05-17 at 12 29 22 PM

BlastyTheDev:
That looks a lot more realistic, the increased number of lines used around the perimeter makes the cloud look more defined and natural

leomcelroy:
I'm going to close this for housekeeping. Feel free to resubmit again in the future.

PR #467, opened 2024-05-13
Cirth-RobertC

nerdwithcomputers:
I made a cirth runes thing, since I was on a tolkien kick when I got the email for this. I've been coding since about 5th grade, mostly in c++, python, and javascript, with some more obscure (and annoying) langs sprinkled in. I _will_ add an update to this that makes it use proper kerning with bt.bounds and things, but I figured I would submit what I had and work on the rest later
anon(1)


leomcelroy:
Very fun project! Thanks for sharing.

Congrats! You've earned a Blot you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw) if you are a teen.

PR #469, opened 2024-05-16
facegenerator-AgentQPK

AgentQPK:
i thought it would be fun to randomly generate wacky faces so i decided to progam a random face generator
 i have 2 days of coding experience in java script and i thought that using blot was easy and fun
veryhappyface


leomcelroy:
I think we may be getting too many of these face generators but this one has some special delightful quality to it.

Can you make sure the iris or mouth get clipped if they overrun their bounding areas (the eye and head respectively)?

leomcelroy:
I'm going to close this for housekeeping. Feel free to resubmit in the future.

Issue #420, opened 2024-04-17
Bt.cover (and bt.cut) render line segment incorrectly when endpoint is close to border


Elijah-Bodden:
When bt.cover or bt.cut is operating on two sets of polylines polylinesA and polylinesB, if any segment of polylinesA has one endpoint inside of polylinesB and one within 0.0001 of an edge of polylinesB (on the inside), the segment will render the opposite of how it should (it will render even though it shouldn't with cover, and won't render even though it should with cut). [Example](https://blot.hackclub.com/editor?shareId=34c2e933-808a-4528-89d3-6a013cd007fa).

Looks like the bug is caused by the following two lines in `pointInPolyline`
```js
// if on edge
const distanceToSegment = pDistance(point, [xj, yj], [xi, yi]);
if (distanceToSegment < 1e-4) return edge;
```
I don't know enough about the codebase to know why being on the lines needs to count as outside instead of inside (or thus how to fix the bug without breaking other stuff) but @leomcelroy probably knows. If this is necessary for other functions, removing it seems to at least not break bt.cover and bt.cut and fixes the bug, so we could maybe add an option that's only true when we call from the boolean function and change the lines to something like
```js
// if on edge
const distanceToSegment = pDistance(point, [xj, yj], [xi, yi]);
if (distanceToSegment < 1e-4 && !ops.reffererIsBooleanFunction) return edge;
```
to disable this code in the context where it causes the bug

Elijah-Bodden:
Okay, patch actually *does* have bad results - when two edges are perfectly overlapping they **both** disappear in cover. I've tried some alternative solutions but haven't found any working solutions without this problem

Elijah-Bodden:
UPDATE: I was wrong, the "bad results" I was seeing were actually my test code being flawed in a way that made the bug *unbreak* it. After fixing the test code, the the bad results with the patch went away. That means the patch (which does resolve the bug) may actually work. Playing around with it, I don't see any bad effects, but there may be edge cases where the tolerance is needed even in cut and cover.

PR #470, opened 2024-05-16
Remove tolerance on pointInPolylines to fix boolean line-type flipping

Elijah-Bodden:
As described in issue #420, checking if a point is within a very small epsilon of of a polyline's border and counting it as outside as so leads to segments with points very close to the border registering as the wrong type (inside gets registered as outside and vice versa) in cutcover ([test case](https://blot.hackclub.com/editor)). Deleting it doesn't seem to have any adverse effects and fixes it.

### DIFF
--- a/src/drawingToolkit/pointInPolylines.js
+++ b/src/drawingToolkit/pointInPolylines.js
@@ -22,8 +22,9 @@ function pointInPolyline(point, polyline, ops = {}) {
     let xj = polyline[j][0], yj = polyline[j][1];
 
     // if on edge
-    const distanceToSegment = pDistance(point, [xj, yj], [xi, yi]);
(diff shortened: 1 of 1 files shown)


Elijah-Bodden:
In case the share link expires, here's the test case:
```js
const width = 125;
const height = 125;

let a = [[[69.99990000000001,65.00000],[57,65]]]
let b = [[[53,70],[70,70],[70,53],[53,53],[53,70]]]

bt.cover(a, b)
bt.join(a, b)

drawLines(a)
```
To clarify this behavior is wrong since because we're using cover we'd expect this line to be covered by the square but it's not because of this bug

leomcelroy:
Let's merge it and see what happens. Looking forward to seeing your town.

PR #471, opened 2024-05-16
Spirilla - Geetha S

geethasreeg:
My artwork basically is a spiral design, which I made by adding the code to form a circle over and over again, until it formed a design like that. It  is a simple design, and was very easy to make. 
image

(I also put in the image before, along with the other files)


leomcelroy:
Thanks for the submission. I think this may be too simple. Try to think of a concept first and maybe sketch it out by hand.

I'm going to close for housekeeping but feel free to resubmit later.

PR #472, opened 2024-05-16
Eclipse-Adam

Adamimoka:
Hello all!
I created this eclipse rendering thing. You can chose the relative sizes of the Sun and Moon, the offset position of the moon, and other color options.

This is my first ever JavaScript code, but I have previously worked with other languages, especially Python.

There is one issue left with with my code, if it is not a partial eclipse, it gives an error.
So that means no annular eclipses (where the moon is completely inside the Sun), no non-eclipses (where the moon is too far), and no complete eclipses (if the moon is larger and completely covering the Sun)

I hope you like my eclipse art.
Thanks!


leomcelroy:
Thanks for the submission. It's a simple piece but it does look nice. Good work with your first JS code.

I think it earns a Blot. You can order one [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw), congrats.

PR #473, opened 2024-05-17
Spirograph-Michael

voidsshadows:
flower


leomcelroy:
I think this may be a bit too simple. What ways could you use a spirograph in a more complex piece?

I'm going to close for housekeeping but you're welcome to resubmit.

PR #474, opened 2024-05-17
shapeGen3D-AliMustafa

alimyust:
My name is Ali Mustafa and my art consists of a class that projects a vertex and edge table to turn it into a 3d shape. This is something I've wanted to experiment with for a while and since I saw HackClub had a program for drawing this type of art, why not do it here? I spent most of my time just learning about the concepts that go behind projection so I would be able to code it, and this is the result.

My program allows you to adjust the shape (from 3 options), the focal length (depth of field), the scale of the drawn shape as well as the rotation on the x, y and z axes.

I've coded in a few different languages (Python, Java, Embedded C, Dart) but I haven't really coded anything for pure visuals before, so this was my attempt.
ShapeGen1
ShapeGen2


leomcelroy:
Nice! You could try to add hidden line culling next.

This earns a blot though. Congrats! You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #479, opened 2024-05-22
TetraTower-Daniel

danielp1218:
My artwork was inspired by Tetris and randomly generates a tower of stacked Tetris blocks. It was both frustrating and fun trying to apply my graph theory knowledge to make this art. In addition, an unexpected amount of coordinate geometry was needed to generate the bounding boxes of the tetris pieces. 
tetra2
tetra3


danielp1218:
An example with colour:
tetra4

PR #486, opened 2024-05-24
art/Turtle-Andrew

Mr-milky-way:
The art is of a Turtle with a shell that is randomly generated (you will have to mess with the seed a lot for it to look good). The outside lines were made in [desmos](https://www.desmos.com/calculator/nvywcy8bam) than put into the JS after making all the ()'s to []'s. I was going to use voronoi noise for the shell but could not figure out how to make one in JS so I used random angles and distance to make it instead. I have little experience with JavaScript, although I do have some familiarity with C#, HTML, and CSS. The image I got inspiration from is just below followed by the image I generated.
print-185176939 (1)
image (6)


Mr-milky-way:
@leomcelroy when you mean different turtles on each run do you mean that the shape of the whole turtle changes? Because the shell changes (you have to change the seed) so do you want it to be more generative?

leomcelroy:
I think it needs to be more different yeah. Perhaps keep the symmetry and change the shape of the turtle, size of flippers and legs, etc...

leomcelroy:
Upon re-review I do think this earns a Blot. I also think you could be improved in the way we discussed. Congrats though you can order your Blot [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #487, opened 2024-05-25
iPhone-Aadarsh

abjectbrake:
My artwork is an iPhone and each time the program is run the size of the phone changes along with the size and position of the secondary camera. There is a slider in the code which can be manipulated to change the height of the camera bar. I am relatively new to coding so I haven't made anything too complex.
snapshot3


abjectbrake:
Okay sure. Thanks for the feedback.

PR #495, opened 2024-05-28
fibonacci-yohance

ypawania:
The Fibonacci sequence - also often called the God's sequence or the golden ratio- gives rise to some of the most beautiful art ever. My code uses three nested for loop to make art using multiple fibonacci squares. It has lots of features that can be changed, including the angle between adjacent squares that make up each complete fibonacci square, increasing which gradually open the square up like a spiral. The distance between adjacent fibonacci squares is also customizable, meanwhile the angle between each whole fibonacci square is randomized to make the image interesting each time. 

I spent considerable time making the fibonacci algorithm work. It took considerable thinking (and rubber duck debugging, pseudocode, and drawing on paper) to make it work. An obvious next step would be to try and replace the nested for loops with a recursive algorithm instead, but at this point I don't see an obvious way to make that work. I had very little javascript experience before this but the syntax was pretty straightforward as blot was very well documented, and there were lots of examples. 

img4


leomcelroy:
Thanks for the submission! I think this code may be a bit brief. I'm going to close this for now but feel free to resubmit when you are ready in the future.

PR #496, opened 2024-05-28
RatandCheese-Anna

AnnaSK7:
My drawing is a rat that, with every run, changes the length of its tail, the angle that the head is looking up, and the overall size of the rat. Additionally, there is a piece of cheese that, with every run, changes how far away it is from the rat and how large it is. I'm kinda familiar with Java, but this was my first time ever doing Javascript, so it was a bit weird--there were parts of it I felt like I should recognize, but ultimately it's a completely new language with completely new mechanics, and I felt pretty out my depth... it's definitely made me interested in what else Javascript can do.

RatSS3


AnnaSK7:
> Thanks for the submission. I don't consider random scales, translations, rotations to be generative variation but I think the rat is quite nice so you've earned your Blot. I'd encourage you to still improve the piece. How can you make the cheese look more cheesey?
> 
> You just need to fix the snapshot names before I merge. `.png` is lowercase and the name has to match what you reference in the metadata.

Thank you very much! How exactly do I fix the snapshot names? Do you want me to just reupload the files with the new names, or is there a way to edit within the pull request?

leomcelroy:
I just fixed it for you. Congrats you can order your Blot [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #497, opened 2024-05-28
pete_the_cat-AlexRabochii

alexfinger21:
Hello, 
I am Alex Rabochii and I drew Pete the cat using blot. I have programmed in JS before, but I have never used the turtle nor the blot frameworks. It was a lot of fun, and I was able to use randomness to create a unique artwork each time. I attached a picture of the art to this PR.
snapshot0


leomcelroy:
Thanks for the submission. Just moving the cat around after the fact doesn't count as generative variation. The cat is well drawn though and you do very some parameters like leg length. I think this earns a Blot.

Congrats you can order your Blot [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #504, opened 2024-05-31
PythagoreanTriples-Mayank

innovativecreations:
I've created a code to draw all the Pythagorean triples whose length of the side does not exceed a curtain length.
For example: If we enter n = 100, then there are 52 Pythagorean triples between 1 - 100, so the code will draw all the 52 triangles.

It's my second PR, the previous one was an abstract art drawing and I love building both scripts.


leomcelroy:
Thanks for submitting again! I'm glad you had fun working on both scripts. I'm going to encourage you to try to create a piece that looks more aesthetically interesting. The pythagorean triple idea could be mathematically interesting but you should try to come up with a visualization of it which also looks nice or reveals some qualities intuitively. Right now it's still a bit haphazard.

Also you still have your old PR on this branch. I'm going to close this for housekeeping but please resubmit again and look for feedback on the #blot channel. You'll get there!

PR #507, opened 2024-05-31
Add files via upload

g1g-1:
girlface-gianna


leomcelroy:
This is really good. I was a bit surprised with the result after reading the code. The girl's face is a bit spooky to me. Not sure if that was intentional but the piece works.

Congrats, you've earned you Blot. You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #508, opened 2024-05-31
Zen-Garden-Satisfying--Arnav Purbiya

APurbiya:
My artwork represents an automatic Zen garden where a stick (on the blot) 
creates satisfying patterns in the sand. Using Blot, I introduced random elements to simulate the natural, unpredictable flow of sand. This experience allowed me to blend coding with creative expression, resulting in a unique, tranquil design.


leomcelroy:
Thanks for the submission. I think this piece is a bit too haphazard. How could you make it more structured.

I'm going to close for housekeeping but feel free to resubmit in the future.

APurbiya:
I thought it kinda made sense as it would make these patterns one at a time in the sand, it would start with the spiral, then the grid and wipe the sand.

PR #509, opened 2024-06-01
DragonCurveFractal-Haneesh

Haneesh000:
I found about Dragon Curves [here](https://en.wikipedia.org/wiki/Dragon_curve). I found them really awesome and amusing. After I found a few fractal arts in the gallery I was determined to make a Dragon Curve. Took me quite a bit of time. I even quit and made a Roblox Logo. Removed the PR because it's not algorithmic. But I was back with this Dragon Curve.
dragonCurve1


Haneesh000:
Also, here is how it works
image

Haneesh000:
Here is how it works: https://upload.wikimedia.org/wikipedia/commons/a/a0/Dragon_Curve_unfolding_zoom_numbered.gif

leomcelroy:
There have been a collection of submissions around fractals that are too similar. Please resubmit with a more unique/original piece. I glad you had fun working on this but we can't accept it for a Blot given the similarity with existing pieces and some inability to determine whether or not that was intentional.

I'm going to close for housekeeping but please resubmit when ready.

Haneesh000:
@leomcelroy I know that there are other fractals, however those fractals completely differ from a Dragon Curve Fractal. The fractals that have been accepted as of yet are infinite image forming fractals. This one is different, it doesn't form a fractal that loops itself on zooming, it is rather a fractal which has been repeated over multiple times, to form this art. I would request you to kindly have a look at the difference and please let me know if this art would qualify. Have a nice day!

Haneesh000:
I just wanted to make it clear that there is no similarity between the Dragon Curve fractal and the other existing fractals. Hope you get my point.

PR #510, opened 2024-06-01
HoneycombFractal-Mohan

vmss2009:
Used the documentation to get to know about the various functions and used some other fractals present in the gallery as examples. Used ChatGPT for suggestions. It took me like 3 days 😅 because of my non artistic experience and poor knowledge in co-ordinate geometry.

snapshot1


leomcelroy:
This seems to borrow from gallery pieces without attribution. I'm going to close it for that reason. It's also too short to qualify for a Blot. Please think of a more unique/original piece.

PR #512, opened 2024-06-01
PizzaMatic-SharonBasovich

sharonbasovich:


leomcelroy:
Thanks for submitting! I think this may be a bit too similar to the other pizza generators in the gallery. Can you think of some ways of differentiating it. Changing the number of slices generated could be one? What else could you do?

sharonbasovich:
Hi, I already implemented the feature to change the number of slices generated as shown in the images in the pull request. Furthermore, my approach is different as it provides greater customizability by making every dimension modifiable. If this is still not enough differentiation let me know what else I can do to satisfy the requirements. 

Thank you,

Sharon

PR #513, opened 2024-06-01
3D-Sphere-Jack

ISaidJak:
When I was younger I would trace my hand and draw curves to make it look 3D. This is inspired by that. I had a lot of fun on this project and I learned a lot.
snapshot1


leomcelroy:
I think this piece is very nice. Simple but pretty. Congrats you've earned your Blot, you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #516, opened 2024-06-01
Planner-Shaheer

TheHackerDude1083:
I had the original idea of making a laptop generator so that's what I started designing. But I've been looking at productivity and I wanted to design a planner. I don't have a printer to make a beautiful planner of my design so this project hits 2 birds in one stone for its design and practicality. This is my first project with drawing other than some small projects with python.
snapshot (1)


leomcelroy:
Thanks for the submission! It's a simple piece but I like that it is functional.

Congrats! I think this earns a Blot. You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #520, opened 2024-06-05
cloudyFibonacci-Yohance

ypawania:
Added complexity and features as suggested by Leo.

The fibonacci square houses a snail with a fibonacci spiral shell. The square floats on water, and there are also clouds in the sky.
There were lots of challenges, even after coming up with the fibonacci square. For one, I tried lots of methods to make the fibonacci spiral, and only after lots of trial and error I found out that I could just make one by making arcs with the diameter of fibonacci numbers. Also, making the lines of the snail natural looking was another challenge, as straight lines did not look very realistic. The skin of the snail had to be noisy and jagged, but also consistent and fairly straight. I tried making my own binomal function by dividing the sum of two random numbers between one and six, but that didn't work out too well and I had to use a boilerplate guassian function. Making the clouds was also very interesting.

Prior to this, I did not have too much experience with art in programming as I mostly worked on python and c++ projects dealing with apis, arduinos, etc.

image


leomcelroy:
Nice job improving it. Congrats on earning your Blot, you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #525, opened 2024-06-07
3d wireframe renderer that takes .stl files

Patcybermind:
its a 3d wireframe renderer!
copy a legacy ascii stl file into the stl variable and it will render it! you can adjust a lot of things afterwards to get the angle right for when you want to draw it!


leomcelroy:
Nice work you've earned your Blot! You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

Patcybermind:
yay!!!! thanks leo!

PR #526, opened 2024-06-08
Cursorwall-smashmaster

javaarchive:
Got a bit busy since I submitted my first attempt, but summer is here so I'm reattempting this as part of hack hour
snapshot1
snapshot2
snapshot3
edit: swapped snapshots with the ones from the download menu

As a background, my original submission might have been a too conservative on blot machine power, so this time I added more smaller shapes and put them in a more intresting pattern. While not visible from this PR a lot of time was spent on figuring out how to calculate angles that lined up with the positions which turned out to be not that hard with a bit of Calculus. 

I think I'd probably want to rename this to "Cursor Wave" but it seems like renaming things on GitHub web is nontrivial.

btw if this is still too abstract please lmk and I can explore other ideas I had in mind =)


leomcelroy:
Thanks for the submission your snapshot reference was off but I just fixed that for you. Congrats I think you've earned your Blot, you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #529, opened 2024-06-08
devilSpider - Eric Hou

yeetric:
My artwork is a spider with devil horns resting on a web.  I have coded in the past but have not used js graphics libraries like turtle so this was a learning experience for me. The eyes body and shadow are made simple jump/arc functions. The legs are made using a function called drawAngledLine which takes in several parameters to draw a 2 angled lines that converge. Then a for loop calls each one a few times, changing parameters slightly to curve around the spider, define number of legs, etc. The horns are created using nurbs and the web is created using concentric circles and spokes within a function drawWeb.  

image (1)


leomcelroy:
This should be approved but you seem to have messed up the PR. I took care of it for you and added it myself.

PR #531, opened 2024-06-09
ProphetOrpheus-Aneesh

Aterani:
My artwork is of the hack club mascot, Prophet Orpheus, sitting down and waving at the viewer. The length of her neck is randomized each time. Her eyes also move with her neck and her head so that her face looks the same every time. I have a little experience coding in Python and Java, but I have never used Javascript before. It was difficult to figure out all the syntax at first, but I got better at it as I worked, and I was surprised that I was able to create this. 
Screenshot 2024-06-08 at 9 45 26 PM


leomcelroy:
Love to see a Blot orpheus. Congrats you can order your Blot [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #533, opened 2024-06-09
Summoning Circles - Armaan Gomes

arghunter:
My artwork is a summoning circle like those described in fantasy novels. I took inspiration from a Banach curve, a fractal my friend told me about and made it look more like a summoning circle. I've done a bit of programming before, but have far less experience than most of my friends. I don't have much experience with JavaScript and also don't really enjoy drawing. Figuring out how to convert a fractal into an algorithm and using recursion to make the code smoother was a rewarding experience. I'd say the greatest challenge was the recursion because I spent a long time trying to figure out why I kept exceeding the call stack limit. I figured it eventually though and like the end result.

image

Here is an image of one of the versions of my artwork.


leomcelroy:
Thanks for the submission! I think this piece is a bit simple. While not a perfect metric aim for at least 150 lines of code.

I'm going to close this for housekeeping but feel free to resubmit again in the future.

PR #534, opened 2024-06-10
StringArt--Ishaan

MatrixMongoose:
This is an algorithm that creates string art for the Blot machine using output from an online [String Art Helper Tool](https://halfmonty.github.io/StringArtGenerator/). Examples from other people were helpful and provided much inspiration. It was interesting building upon the code of someone else in a cumulative effort. 

image


leomcelroy:
Thanks for the submission, I love the idea.

Could you write the main algorithm yourself? I think this relies on an external tool too much.

PR #536, opened 2024-06-13
WavesHorizon-AudreyWang

audreyolafz:
I was inspired by waves, whether that is looking into an ocean or a mountain, so I wanted to create waves that were randomized. The flowers are drawn algorithmically. This is my first time working on generative art so I'm pretty happy with how it turned out.

snapshot1


leomcelroy:
Thanks for the submission! I think this earns a Blot. Congrats you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #537, opened 2024-06-14
Fractal-Mohit

atomisadev:
This abstract fractal framework was generated using the Blot library. The fractal tree is generated recursively, with each branch splitting into two smaller branches at random angles. The tree is drawn vertically from the bottom up. The `fractalBranch` function recursively draws the tree branches. It takes in a turtle object, branch length, angle range, and length decrease factors as params. If the length is greater than two, it draws a branch, creates left and right sub-branches by copying and rotating the turtle, and recursively calls itself with decreased length.


leomcelroy:
Thanks for the submission. I think this piece could be a bit more unique, it's a pretty standard fractal. 

Keep at it! I'm going to close this for housekeeping but please resubmit again later.

PR #540, opened 2024-06-16
HotHotRacinCar-NoahCarmichael

noah-carmichael:
snapshot1

This is my blot artwork program. I made a 2d racing car and a simple backdrop with a few generative variables:
- Spoiler angle
- Roof type (blocky vs smooth)
- Rim art shape
- Racing stripe pattern
- Cloud location
- Cloud shape (very subtle changes)

I tried to recreate the style of car I used to love drawing when I was a lot younger. I used a combination of lines and the turtle program to make the picture. I modified the Catmull-Rom curves code from one of the workshops to give the racing stripe unique patterns. Overall drawing with "code" seems way less straight forward than by hand, but the ability to automatically modify drawings and control how they're modified was very interesting to work with.


leomcelroy:
Nice! You've earned your Blot, you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #543, opened 2024-06-17
UFO-Sajeg

Sajeg:
This was a lot of fun creating. It's a Unknown Flying Object. Also known as UFO. It sometimes appears in Sci-Fi movies and some people are convinced, that they saw one in real life. Often they are just satellites. But maybe, just maybe there is a little truth behind it. Because I wasn't sure how it looks I made some parameters changeable, so that everyone can create their own version of this UFO.
I have some experience with coding, however not with Java Script. I hope that this artwork is eligible for the cool blot. 

image


leomcelroy:
I think you are on to something but this piece is a bit too simple. While not a perfect metric aim for around 150 lines of code.

I'm going to close this for housekeeping but feel free to resubmit again later.

PR #546, opened 2024-06-18
boatart-bhattprayas

bhattprayas:
I created a boat with some waves. Coding with blot was a very good experience and it helped me to learn more things.
image


leomcelroy:
Thanks for the submission. I like the simple aesthetic but this piece may be a bit too simple.

How could you make it more sophisticated while keeping the elegant look?

PR #550, opened 2024-06-18
collatzVisualization-kate

katexu66:
Artwork: a visualization of the Collatz conjecture, creating a coral/bird-like structure using an initial number and certain number of branches. I find the intersection of art and math/science beautiful :) It's amazing that a simple algorithm of two operations and a program can create such organic, intricate structures.

Experience coding: very experienced with Python, not so much with Javascript! So this was a really fun project to experience Javascript and see what I could create with a language I'm not as familiar with.

snapshot1


leomcelroy:
Thanks for the submission. I think you are on to something but right now this is more of a math demo. It evokes reeds or grass to me. How could you make the piece more figurative?

PR #552, opened 2024-06-18
simply (the best) skyline - Marco Samuele BELOTTI

marco-belotti:
it's an easy skyline generator, you put in number const the number*2 of palace you want and click start!


leomcelroy:
Thanks for the submission. I think this piece needs a bit more refinement. There are also a few errors in your submission (snapshots folder and naming). I like the perspective but it doesn't quite evoke buildings, it feels more like a math graphing demo now.

I'm going to close this for housekeeping but please resubmit after refining the piece.

PR #554, opened 2024-06-18
FractalHarmony-AlexB

oldfashionedjelly:
This project was a lot harder than I expected it to be and it took a lot longer than I thought, but I'm glad I stuck to it. After a lot of trial and error and many failed ideas, I decided to go with a fractal design because it could be done recursively with relatively little code and it would allow for a lot of variation by adjusting the parameters. The artwork itself features a hexagon filled with different fractal shapes and designs. Things such as the size, complexity, and number of layers can be adjusted. Overall, I'm happy with how it turned out and that I was able to make something cool, despite any failed attempts initially.

image1


leomcelroy:
Thanks for the submission! I think this is a bit too chaotic. I would recommend thinking of a new concept for a piece.

PR #556, opened 2024-06-19
DragonBall-Mayank

innovativecreations:
I grew up watching Dragon Ball, so I thought of writing a script to draw Dragon Balls.
Building the code is gratifying, especially when it doesn't work as it should. Although it's my 3rd third idea I've enjoyed each building process 
snapshot1
snapshot2
snapshot 3

Thanks, Hack Club, for such an experience


leomcelroy:
Thanks for being persistent. I think we discussed this on Slack that you were going to try to make the dragon balls match the canon. I think this is close but not quite right.

leomcelroy:

PR #559, opened 2024-06-19
randSmileyGen-shbPng

shb-png:
My artwork generates a smiley each time you run the page. The eye location/size as well as mouth orientation/size are determined by random number generators. I mainly have experience in web development (basic HTML/CSS, but I am beginning to learn how to integrate JS into the whole thing). I also have learned python as well as teeny bits of C++ for competitive coding challenges.


leomcelroy:
Thanks for the submission. I like this but I think it's a bit too simple. Especially considering there are a few smiley face generators already. How could you make it more unique.

PR #560, opened 2024-06-20
retroSunset - Luke Thomas

Zoobdude:
snapshot1

Inspired by retro sunsets like this one:
image

I have exposed many parameters at the top of the script to customise:
* The width of the rays
* The thickness of the border around the sun
* The distance of the rays from the sun
* Various colours

Some parameters are random too.

I realise that the colours would not be visible with the blot, but I think it is still effective without the colours.

This is my first time doing a proper turtle project so I'm sure there are more efficient methods of doing things.

I hope you like it!


leomcelroy:
Nice looking piece. Congrats you've earned a Blot, you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #562, opened 2024-06-20
euFlag-opensource25

opensource25:
The artwork is a canvas with a variable number of stars of a variable size arranged in a circle around the center. It was programmed to create the flag of the European Union after I saw a US flag in the gallery.
I've never programmed anything like this before and not much in JavaScript, but programming after school was really fun.
snapshot1


leomcelroy:
Thanks for the submission! I think this piece may be a bit too simple. I'm going to close this for housekeeping but please resubmit when you are ready.

PR #563, opened 2024-06-20
SunlitCastle-Alvin

ThunderEdge73:
The artwork is of a castle with a clear, sunny background. This artwork is similar to what I would have produced if I were told to draw a castle on a sunny day, and I am happy that I could program something to draw something just as detailed as I could. I included some randomization to the position of the sun, the number of bars (horizontally and vertically) on the portcullis, and the shape of the top bits of the castle.
I've had 3 semesters worth of coding experience in my high school, but most of the time in those classes was spent learning a little bit of Scratch and mostly Python. I had to search up the basic syntax for Javascript basics more times than I would like to admit, but I found that a lot of the essentials I learned from Python (sequencing of events, using and creating functions, etc.) transferred to this project, making this a surprisingly fun experience.
castle1


leomcelroy:
I like the hand drawn feeling it has. Congrats you've earned a Blot. You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #564, opened 2024-06-20
Add The Race art

polypixeldev:
Adding "The Race", my first complete Blot program! The art shows a top-down view of a race with two lanes and a variable number of cars. I used a reference image for the outline of the car design, but made my own changes where I thought it would look good, and the design in the middle with the noise is something I designed.

I had a really good time making The Race. I don't consider myself much of an art-sy person, but I found it surprisingly fun to be able to describe a scene in JS in terms of polylines and turtles. I also wasn't expecting myself to be thinking of a lot of math and trig over the summer, but I actually liked being able to put the concepts I learned this past school year to use in a real program.

I logged around 6 hours working on this with Arcade, but I've spent a fair bit of time before this art learning how Blot works and going through the guides. One of the concepts that I understand better now after using it in The Race is noise, which didn't really make all that much sense to me before.

Screenshot from 2024-06-20 16-44-47


leomcelroy:
This looks great. Congrats you've earned your Blot. You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #565, opened 2024-06-21
hackclubblotproj-amanda

amandasff:
program generates various slightly different iterations of the letters HC within a box.
image

experience coding: have taken some courses in and outside of highschool.


leomcelroy:
Thanks for the submission. I think this piece could be refined a bit more. Also pay attention to the file structure of your submission. You've got two folders with a few issues in each right now. Keep at it! It's a good start.

I'm going to close for housekeeping but please resubmit when you are ready.

PR #566, opened 2024-06-22
Spiral Shapes-Max

maxdrapa:
I have very little experience coding so I was using this as a way to get a bit more familiar as well as get something out of it. I decided to make a spiral because today I threw away a school paper today and saw polar spirals on it. With those fresh on my mind I wanted to make some with blot and I wanted to make it a bit more special by adding shapes of any size to form the spiral.
image


leomcelroy:
Thanks for the submission! I think this piece is a bit too simple to earn a Blot. I'm going to close this for housekeeping but you can resubmit again later.

PR #567, opened 2024-06-22
convergence-sashmitAarya

Sashmit29:
screenshot1 png

This is an art piece depicting a railroad, drawn with the 'one-point perspective' drawing technique. This art technique uses a 'vanishing point' on the horizon line of the drawing to create a comprehensive illusion of three dimensions on a two-dimensional surface. The 'vanishing point' is a point on the drawing where all parallel and receding lines in the drawing appear to meet, or converge. That is what inspired the name of this art piece.

There are also extra 3D effects added to the horizontal tracks and the vertical steel rails of the railroad to make it seem as if they have height (thickness) and width, instead of just being a plain line.
Furthermore, to add even more 3D effect, almost all aspects of the drawing get scaled down as they get "farther" from the viewer. These include:
- The space between consecutive tracks
- The thickness (height) of the tracks
- The width of the tracks
- The width of the steel rail

There are four adjustable parameters located at the top of the program: the bottom railroad width, top railroad width, railroad height, and number of tracks. The top railroad width should always be 0, because it is the 'vanishing point'. The bottom railroad width and railroad height can be scaled as the viewer wants. The number of tracks is the random element in the art piece. There is an algorithm that I had to figure out, that allows for any number of train tracks to be displayed while still having the 3D effect of exponentially scaling down the distance between consecutive tracks as they get "farther" from the viewer.

I have about 3 years of experience in software engineering. My coding experience is mainly in Java and Python; however, I did have a fair amount of experience with JavaScript before starting this project.


leomcelroy:
Well done! It strongly evokes tracks, has a nice mathematical element, and is pretty.

PR #570, opened 2024-06-22
Imperfections-Dhruv

ThinkerDesigns:
Life can't always be perfect. This can be seen in my artwork where all the squares look like they have perfect lines with each other but right in the middle (climax) there is an irregular line that disrupts the flow of the squares before they resume the regular flow. My experience coding with not just this website but other languages has been quite engaging and I plan to major in computer science when I do my Common App this fall.
Snapshot1


leomcelroy:
I think this piece is a bit too simple. You can definitely do interesting things with variations on grids though. You can resubmit when you are ready, thanks for getting into Blot!

PR #572, opened 2024-06-22
FramedChildishDrawing-ValeriaDA

vale217:
Hello!! I decided to make a drawing inspired by the type of drawings kids make and that get hung up on the walls to motivate them. Instead of just making the house, smoke and hills, I wanted to put it in the context of it being inside the frame tied to a wall with a ribbon. 

I am a tutor that works with young students so I wanted to create something simple that reflects that and that would still allow me to get a hang of coding as I am completely new to it. It is a goal of mine to make progress during the summer with my free time and I feel this is a good first step. Thank you :)

snapshot1


leomcelroy:
Thanks for the submission. Could you add more variation to the image on generation? Right now only a few of the smoke lines get modified.

vale217:
Thanks for reviewing my submission. I was able to edit the "index.js" file with the changes I have now made but I am not able to do the same for the snapshots so here they are:

snapshot1
snapshot2
snapshot3

leomcelroy:
Nice.

PR #573, opened 2024-06-22
Staring at the Tree - Eesha

EeshaV28:
This artwork looks really simple, it is just a tree that is randomly generated, along with a person looking at the tree which is also randomly generated. I am not that good at JavaScript because I primarily use Python. Most of my time spent working on this was spent reading through the documentation to get used to polylines and the different ways to generate curves.
screenshot1


EeshaV28:
Most of my time spent working on this was reading through the documentation

leomcelroy:
I think this needs some more refinement before earning a Blot. How could you add more detail to the tree or person? Some generations completely flatten the person now.

PR #577, opened 2024-06-23
AFlowerForYou-SidakSingh

Sidak08:
My artwork is a piece of code that generates a special flower for you. A unique one out of 80028000; the one that is made just for you. I started this idea out with the plan of it being a Mother's Day gift, but after checking the date It is a bit early. But that does matter. Let's talk about something that does matter. My experience building it. I would say I had a mostly great experience building it, except for the part where I just didn't think to consider that there could a docs for blot and made my way through mostly using trial and error and my beloved console log. Furthermore, even after discovering the docs, I didn't find anything about making a circle which was kind of weird; but I got a refresh my trigonometry skills. Except that, I would say I had a fun time coding after a long while. As I didn't get to code for the past 2 months because I had been extremely busy with school and my exams. The last time I codded was at a hackathon, I don't even remember the last time I even worked on a passion project. 

image2


leomcelroy:
Could you use cut/cover so there aren't so many lines overlapping. I'm going to close this for housekeeping but please resubmit when you are ready.

Sidak08:
Hi I resubmitted it with less overlapping lines. https://github.com/hackclub/blot/pull/596

PR #579, opened 2024-06-24
RainWeatherSimulator-EthanJohn

Somebud0180:
Whilst brainstorming, one of the shapes I could form was that of a cartoonish raindrop. I went ahead transformed that to a randomly generated raindrop. After awhile I figured I needed a cloud, which was added later. I had a bit of difficulties with coding as seen with the tendency of the raindrops to overlap, it was outside my skillset to make something that avoids that. The drawing part I really learned as I went, and specially with the help of the community it was made easier.
snapshot1


leomcelroy:
Nice. It's a simple piece but I think you captured the cloud well. You've earned a Blot.

It would be good if you made the clouds change too. Also you can use `cut` and `cover` to get the raindrops to occlude each other.

Congrats you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

Somebud0180:
Just updated this, took awhile since I got stuck and moved onto Sprig. 

- This adds new 3 new clouds which are randomly picked to spawn (or not)
- Raindrops don't overlap anymore and are more evenly spaced out
- And a few changes here and there

PR #580, opened 2024-06-24
RandomTreeGen-Aradhya

Aradhya-Shaswat:
This artwork is a tree upside down with leaf (represented as circles!) and it is randomly generated by a randSeedGen. I spent around 6 hours coding this and going through the docs and I am happy with the results. Overall, I liked this project a lot and I would love to get a Blot to try this out irl! Thanks, aradhya


leomcelroy:
Thanks for the submission. I think this may be too simple to earn a Blot. Could you add some thickness to the branches and make it look more natural?

PR #584, opened 2024-06-25
SeasideSerenity-Shreyas

shrysjain:
My artwork is an ocean scene consisting of many trigonometric and randomized patterns. The main features of the scene are the waves, the sun, and the clouds, each of which have variables which can be modified or randomized. For the waves, the frequency, phase shift, number of waves, and interval between the y-coordinates of each wave can be found in variables at the top and randomized with a `randInRange(min, max)` helper function which I wrote. For the sun, you can randomize the radius, and for the clouds you can randomize the size of each circle creating a cloud. The waves are sine waves drawn with the function `startY + 3 * Math.sin(frequency * (x - startX) + phaseShift)`. I had a lot of fun working on this project and I can't wait to build a Blot machine and have my work drawn :)


leomcelroy:
Thanks for the submission. I think each element is a bit too simple. How could you give the clouds more texture without relying on fills? Could you create the effect of light from the sun?

PR #585, opened 2024-06-26
Snowflake - HenHen

46-HA:
I made snowflakes for blot! They are randomized so each snowflake you generate is completely new (just like in real life!)
Check it out, I think it's super cool. 
Snapshot1


leomcelroy:
Thanks for the submission. I'm not sure this quite captures the look of a snowflake. I'm going to close this for housekeeping but please resubmit when you are ready.

PR #588, opened 2024-06-26
2PointPerspectiveArt-Henry

Henry-Moffatt:
A 2 point perspective art, with 2 functions, "loss" and "skyline". Skyline creates a city skyline at night with stars, though the stars only show in the preview and not the final export. Loss recreates the loss comic in 2 point perspective. Limited experience in python and C#, no experience in javascript, though this project required a lot of maths and linear algebra and trigonometry rather than extensive coding. Thank you hackclub.

snapshot


leomcelroy:
I think this piece is simple and effective. You rely on fill to cover things up. Could you use cut/cover?

I think this earns a Blot though. Congrats you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #590, opened 2024-06-26
SquareCircle-AdamMiller

AdamCodes2023:
This is my Art Program Submission to receive a Blot Robot. My piece is named "Square Circle" because the center of it is dominated by a revolved square that looks like a circle. There are also several revolved triangles to add angelic shapes throughout the piece. To add a little bit of randomness, a curve is randomly generated across the boundaries of the piece as a final touch. I have never programmed art before, so this was a very enlightening experience. I enjoyed using my technical knowledge to create real art that I do not have the abilities to create by hand. Thank you for this opportunity, and I am excited to see my Blot in action.
- Adam Miller :)

Snapshot_1


leomcelroy:
Thanks for the submission. I think this may be a bit disorder to earn a Blot but please resubmit when you are ready.

PR #591, opened 2024-06-26
PyroSnow-Map

GGGalang:
# ABOUT:
**_Pyro Snow: A World of Black and White_**

I heard from a YouTube video once that snowflakes were "completely unique" from one another. To challenge this statement, I created what I call a simple generator of such snow, with the ability to generate a set of a random kind of snowflake every time and display 30 on screen at random coordinates under a dark cloud. While obviously borne from illusions, see the different snowflakes right in front of you; the program has pentagons, heptagons, square-agons(?) and much more!

Quite honestly, I was surprised myself that such complex drawings resulted in illusions that highly resembled snowflakes. While the highly illusory effect was unintentional, the outcome is; so regardless of your country being in the equator or not, please enjoy the snow!

Originally, the colors were supposed to be inverted for an extra feeling of being unable to touch the snowflakes on zoom; then I realized the Blot machine probably can't handle it, and I was forced to return it to this color scheme. How unfortunate!

# CODING EXPERIENCE:
I commonly use _HTML, JS, and CSS_ (while I know they aren't exactly programming languages) to build websites. I also have experience in other languages such as Python and Java. While I have experience tinkering with JS to make website elements do wonders, it's only in Hack Club where I ever found myself doing these unexplainable kinds of things. Blot is quite challenging, especially on JavaScript! But it was quite fun getting to where the art is at right now, too! I never found myself to be an artsy type so this was quite an experience.

snapshot1
snapshot2


leomcelroy:
I think you could improve the cloud. How could you give it some texture?

I think this earns a Blot though. Congrats you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

GGGalang:
Hmm I think I can probably create better clouds with more curves! I'll try revisit this project soon and post in the channel if I achieve anything

Thank you very much!

PR #592, opened 2024-06-26
Star Flag

ridwann89:
So i did my project with Javascript for the coming up 
anniversary where Somalia declared independendce from Italy in the 1. of July 1960.


leomcelroy:
Thanks for the submission! I think this needs a bit more sophistication to earn a Blot but please resubmit when you are ready.

PR #594, opened 2024-06-27
WoodOnFire-AndrewCromar

AndrewCromar:
My artwork started as just an idea, to see if I could draw a flame of fire. Little did I know after adding the wooden background, the whole thing went up in flames.

At the top there are many variable to customize the randomness of everything.

My experience with this project was fun. I had to learn how to use the line drawers, I had previously used js for discord bots and website but I have never done anything like this.

snapshot


leomcelroy:
Thanks for the submission. This relies on fills so it won't draw that nicely with a Blot but you used the colors well to make the flames. I think you've earned a Blot. Congrats you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

AndrewCromar:
> Thanks for the submission. This relies on fills so it won't draw that nicely with a Blot but you used the colors well to make the flames. I think you've earned a Blot. Congrats you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

Thank you, I knew it would not draw correctly but I thought it would look better.

PR #597, opened 2024-06-28
DottedGrid-Rushil Chopra

WolfGamer2:
This is my second try! I hope
snapshot0
 this gets accepted! I worked very hard on this :)


leomcelroy:
The submission criteria suggests making a piece that depicts a real life thing. I think this is still very abstract. Try to think of a concept that matches the submission criteria.

PR #598, opened 2024-06-28
Mandelbrot set - Pere

Pegoku:
I love fractals, and wanted to create the mandelbrot set with blot. It wasn't easy but i finished it and it works, you can stretch it, change size, details, etc.
image


leomcelroy:
This is very similar to an existing piece. Try to come up with an original concept.

Thanks for submitting though and please submit again when you are ready.

Pegoku:
Thanks for answering. 
Also, would like to know if a svg to blot script would be an adequate piece. 
Thanks

PR #599, opened 2024-06-28
Since my last code got rejected (rightfully) i present a completely diffrent thing that i think is much much better :D

saumilthecode:


leomcelroy:
Thanks for resubmitting. I think you could take more care with this piece. What are you trying to depict? The image still seems very abstract.

Take your time with resubmitting to carefully craft a piece. Feel free to seek feedback on the Blot channel.

PR #601, opened 2024-06-28
Resubmission UFO by Sajeg

Sajeg:
Hey thanks for looking at my past PR. I could understand your feedback.

Even-though it's still not 150 lines of Code, I think it looks way better. And I also don't have any ideas anymore of what I could add. The Math behind the UFO took me a while to figure out and that's one of the reasons why I think it's not too bad that I don't have 150 lines. Another reason is that I always tries to make my code very compact. I could've just stored every equation in it's own variable, but I thought that it's then to annoying to read.

So I hope you'll accept my PR this time and even when you deny it, I still thank you for taking your time and looking at this again.


leomcelroy:
I think you could still do more to improve this. Take your time if resubmitting.

PR #602, opened 2024-06-28
Shrek-AnirudhPaladugula

PRODOFFICAL:
The coding of Shrek was a decent challenge for me and it involved a lot of me tweeking my code back and forth each time to make an ajustement to the code for a better looking image 
Shrek3


leomcelroy:
Thanks for the submission! We have a few faces like this already in the gallery so you have to put in more effort with new similar submissions. 

A few things I'm curious about.

- Why is it upside down?
- Why is it drawn multiple times?
- Could you use more than basic geometric shapes?

I'm going to close this for housekeeping. But please submit again after you improve your piece.

PR #605, opened 2024-06-28
Fix guide+shareId links

### LINKED ISSUE #603: Share link broken when shared from guide
When you create a share link while on a guide, the share link won't work. 
Fixes:
1.  make it so all share links don't have the guide paramater in the URL
2. make it so even if it does have a guide parameter, it will still load the shared code

fix 1 is easy so i'm going to try and figure out fix 2 in the meantime

BrightTheBackpack:
closes #603 
If you're interested in fix 1 lemme know. also, cause my enviroment doesn't have access to the database and stuff, i'm only 90% sure this works.


leomcelroy:
Thanks Bright. Could you remove the nix stuff and the package updates?

BrightTheBackpack:
all done!

leomcelroy:
Nice catch finding this.

leomcelroy:
Thanks for pointing this out. I think the more general way to fix this is to strip URL params when generating the share link. I did that in this commit `dc8d82a0`.

I'm going to close this PR but you deserve credit for catching it.

leomcelroy:
If you realize I didn't account for it properly let us know and we can reopen.

PR #606, opened 2024-06-28
BeautifulSaturn-AlexB

oldfashionedjelly:
This was a lot harder than I expected, but I'm happy with how it turned out. It took a lot longer than I expected and it pushed me out of my comfort zone with using JavaScript and Blot. However, I'm happy that I put in all of the time that I did. There was an immense amount of trial and error required as I went through at least five different design ideas, but I chose this one because it is simple yet artistic and can be made pretty easily with the machine. It features the planet Saturn with some stars in the background. You can also change things like the size and angle of the planet and rings and the number and size of the stars.

snapshot1
snapshot2


leomcelroy:
This is almost there!

Do you think you could have the back part of the ring covered by the planet?

Also make sure the snapshots folder is all lowercase.

maxwofford (changes requested):
ooooh, out of this world! Yeah, if you fix the things that @leomcelroy [mentioned](https://github.com/hackclub/blot/pull/606#issuecomment-2216304052) I can merge this

oldfashionedjelly:
snapshot3

I finished making all of the changes!

maxwofford (changes requested):
Nice work! A couple other things come to mind though– the stars show through the rings. Also, in most photos of saturn, it looks like there are a couple rings. What if you add that in?

Screenshot 2024-07-14 at 03 18 17

oldfashionedjelly:
snapshot1

How's this?

maxwofford:
Lol, [that's one way](https://github.com/hackclub/blot/pull/606/commits/e09a591b788fe3db11f709ec0a1926647852dbe0#diff-9f0cf74f710b390ab3e7f33e117f0bffab5631bfe342483230f6f908a957d0f1R79) to do it

maxwofford (changes requested):
It's looking really nice now! Almost ready for merge. One last thing: it's not quite generative enough– I tried rendering it a couple times and only the stars move their positions and counts. A couple things you could try out:

- variable number of rings each render
- variable ring size
- variable ring rotation (oooh, this would look dope)
- different star size (smaller stars could look further away & form constellations)
- different planet size

Once you have a couple of those things in this art deserves a blot!

oldfashionedjelly:
I fixed the star placement, scaled the planet and ring size by a random factor each time, made the ring rotation vary, and made different star sizes.

maxwofford (changes requested):
This looks fantastic! Unfortunately it's still running outside the (blue) bounding box. Fix that so I can merge!

oldfashionedjelly:
It should be fixed now!

maxwofford (approved):
Absolutely beautiful work on this! It def deserves a blot. Go ahead and [order yours now](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw)!

PR #609, opened 2024-06-29
Generic 3D Bed Slinger - JustJoshingYa

JustJoshingYa:
Hello!
This piece of art is a non specific 3D printer. It was inspired by the Bambu A1 Mini and the Prusa Mini. 

The X and Z axis are interactable via the commented out variables. By default, the X and Z axis are randomly set to different positions. There are also limits to the Z and X axis to prevent the toolhead from coming off the 3D printer.  There is some filament that follows the toolhead's movement as well.

The screen is also interactable with a commented out variable. Currently, it is also randomly set between 3 images and 1 blank screen. The images present are The Blot logo, the Klipper logo, and some text that says Calibrating. Each one of the svgs was free and open source or generated. Links have been commented in as sources.

Coding this was hard without the turtle. When I learned how to use the turtle class, it was just a matter of designing the printer. The Nurbs curve filament was interesting to code as I wanted a solid line of filament to travel to the toolhead, but the toolhead was always moving. Thus learning how to use the Nurbs curve was very useful. Halfway through the 12 hours I used to code this, I realized the code base was JS script which made the filament path possible as I could use for loops to my advantage.  

I also tried to add a printer within the printer and make it look like the printer was printing a printer, but that required filtering through thousands of polylines to  and subtracting select points on the y axis to make that happen so that feature was scrapped.


leomcelroy:
Great piece! Some of the screen text appears to be mirrored.

You've definitely earned a Blot though. Congrats you can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #611, opened 2024-06-29
Infinite-Skyline-Rishabh

Rishabh416:
My artwork is inspired by a simplified view from my favorite category of images, skylines. I had an amazing time coding this project and using blot to be able to create my own skyline drawing. With this code you are able to configure almost every property of the buildings by modifying the set of variables at the top of the file, giving you control over your favorite skyline. I liked how it turned out, especially with the addition of the colors (took me longer than it should have but got there in the end) since it really makes it look like the buildings I see around where I reside.


leomcelroy:
Thanks for the submission! I think this piece may be a bit too simple. Could you add some perspective perhaps?

I'm going to close this for housekeeping but please resubmit in the future.

PR #612, opened 2024-06-29
NoOverlappingBalls-Krish

ikrishagarwal:
hello!!

This code uses the mathematical formula of calculating two circles overlap or not and thus draws only those circles which would never overlap with any other circle.

I derived this formula myself even before the school taught us. I finally got a use of those old works and turned them into code :)

To end, this I loved learning about canvas drawing with blot and enjoyed integrating maths with programming, a big thanks to the community ❤️


leomcelroy:
Thanks for the submission. This piece is a bit short but it's a nice example of packing (of some sort) for the gallery. 

PR #613, opened 2024-06-29
flagGenerator-AlbertCoolGuy

AlbertCoolGuy:
My artwork is a flag generator. It generates a flag with 3 colors and chooses a random flag design.

image


leomcelroy:
I think this is a bit too simple to earn a Blot. Thank you for submitting though!

I'm going to close this for housekeeping but please submit again when you develop a new piece.

PR #614, opened 2024-06-29
RobotClipart-Krishnan

krishnans2006:
This is some clip art of a robot's face. Last week, I saw a similar design of a robot face hung up on a wall, and thought it looked very cool and satisfying. This is my attempt at recreating it.

My experience has been very fun - despite a lot of tedious number calculations which I'm sure I could've abstracted, seeing each part fit together was so cool (and there were clear indicators of progress to keep me from boredom).

image2


leomcelroy:
It's a simple piece but a nice variation on the face generators we have in the gallery.

PR #615, opened 2024-06-30
DontLookattheSun-Benjamin

53142:
My artwork depicts multiply green, eye-like figures seemingly looking at a red sun in the middle, hence the name of the artwork: _Don't Look at the Sun_. I don't have that much experience in programming, however, I have some familiarity in designing websites with HTML, CSS, and JS. Creating this artwork was a fun challenge. There was a lot of tweaking involved with trying to line everything up in the correct spots, but, overall, I like how my artwork turned out.


leomcelroy:
Thanks for the submission!

I think this may still be a bit too abstract. I'm going to close it for housekeeping but feel free to resubmit again in the future.

PR #616, opened 2024-06-30
PartialHatsScenery-Arnav

arnav-kr:
I saw the [Einstein tile](https://en.wikipedia.org/wiki/Einstein_problem) when its solution was found in 2023, and wanted to try making something with these tiles, in the following blot, I have attempted to visualize an abstract scenery using a mosaic of Einstein tiles, took references from the original published paper [here](https://arxiv.org/abs/2303.10798),

here are some of the snapshots of the project


leomcelroy:
Nice to add more tiling patterns to the gallery. Could you add variation by changing the shapes of the tiles on each run?

arnav-kr:
Thanks!

um, those shape are its identify 😅, it is a shape that can tessellate space but only in a nonperiodic way

leomcelroy:
I understand. I mean to create other tiling patterns (periodic). 

This is an interesting one to draw peoples attention to for sure.

PR #619, opened 2024-06-30
CollatzKite-Kate

katexu66:
This is a line art drawing of a kite, with flying kite strings generated by a Collatz conjecture visualization. The number of kite strings and initial number used to generate the strings can be changed. The kite body also has dot shading along one side.

I have experience with Python (have done a few projects using Python before), but this was one of my first using Javascript! I have previously used Javascript, but only for making a few easy games, and it has been a while, so this was a really nice review/opportunity to learn more.

snapshot1


leomcelroy:
Nice job using a math concept in an image!

PR #621, opened 2024-07-01
MessageMaze-Iambodha

iambodha:
# Message Maze

Message Maze is a maze creation tool that allows you to embed hidden messages within mazes. And then draw them via Blot!

## Features
- **Hidden Messages**: Create mazes with concealed messages.
- **Python Logic**: Maze creation logic implemented in Python due to familarity.
- **Open Source**: The code for the API, website, and maze creation logic is open source and available on [GitHub](https://github.com/iambodha/Message-Maze-Blot).
- **Try It Out**: Visit the [Message Maze Website](https://iambodha.github.io/Message-Maze-Blot/) to create your own mazes with hidden messages.

## About
I thoroughly enjoyed coding this project—it was a lot of fun! I am hopeful that it will be accepted into the gallery. I look forward to receiving a blot.

## Example
Attached is an image of a maze with the hidden message "hello world". I believe such mazes would make excellent message cards for gifts.

HelloWorld-1

### DIFF
--- a/art/MessageMaze-Iambodha/index.js
+++ b/art/MessageMaze-Iambodha/index.js
@@ -706,6 +706,10 @@ function drawMazeFromString(hiddenMessage) {
     const mazeData = generateMazeListFromString(hiddenMessage);
     console.log(mazeData);
 
+    let lastElementMainList = mazeData[mazeData.length - 1];
+    let lastElementSublist = lastElementMainList[lastElementMainList.length - 1];
+    lastElementSublist[1] = 0;
+
     // Set dimensions
     const rows = mazeData.length;
     const cols = mazeData[0].length;


leomcelroy:
This is a great project. I see the core of it also seems to have been written by you. Could you include that in the submission? Ideally it would be done in your Blot project. ChatGPT could probably help you port it over pretty quick.

maxwofford (changes requested):
Woah, this is sick! Great work on this. Could you rename your snapshots to be lowercase so I can merge this PR?

iambodha:
I am not sure if this is what i was supposed to do, but I think I just made a new request with the changes. Sorry for the incovienence if i did it wrong.

maxwofford (changes requested):
Oh, looking at the code, it looks like this is generated in python & then you're pasting the output into the blot editor. You should rewrite this in JS so others can experience the generator you made while they're browsing the gallery!

The editor def has enough processing to handle it & I know rewriting python -> js can be a learning curve, but it'll be a great way to learn how to build stuff in the browser in the future!

maxwofford:
Just posting this as an update– we have a discussion right now on the team about how to handle async code. Sorry for the wait!

BrightTheBackpack:
You can use xmlhttprequests for api calls as it's non async

iambodha:
I have converted most of the python logic to java script and now only use api request to get the alphabet patterns in the editor(json files on my github). I hope API Requests for those are fine.

The new problem is that the blot editor is either timing out, running out of ram(500mb according to chrome) or getting stuck somewhere along the script. My script works fine on local machines as the new server side logic gives the desired output and when I copy this output and put it in the Blot editor it works fine. So I am not sure where the problem is. If you could help me find out the problem it would be awesome. 

Here is the link to the new script: https://github.com/iambodha/Message-Maze-Blot/blob/main/Updated-MessageMaze/index.js

iambodha:
I looked futher into it and the script actually does get a list as the output but for some reason the list does not work. If i would copy this list and paste it in it works. I think there is a problem with the await or something else in the function.

iambodha:
The problem has been solved, on the normal blot editor async is not available but when a temporary blot editor with async was used it worked perfectly fine. I hope this pull request still gets accepted :)

BrightTheBackpack:
image

qcoral (changes requested):
I LOVE this!!! Definitely approved. One thing before I merge though - there seems to be a bug with the new javascript code, as it doesn't seem to have a solution! If you can fix that, I'll merge this and send you an order form!

qcoral (approved):
Thanks for making the fixes! You are now approved for a Blot. You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw)

PR #627, opened 2024-07-03
CatFaceGenerator-Cral-Cactus

Cral-Cactus:
In my artwork, I've coded a whimsical cat face generator using procedural drawing techniques. The code defines functions for drawing various elements such as ears, eyes, mouth, nose, and moustaches, each with randomized styles for a playful and unique appearance. It incorporates bezier curve transformations and noise functions to add smooth transitions and organic textures, resulting in a lively and interactive visual experience.


leomcelroy:
Thank you for the submission! I think this may be too similar to the face generators we currently have in the gallery. How could you make it more cat like? Perhaps draw a cat body too.

leomcelroy:

PR #628, opened 2024-07-03
hackclubServer-Joseph

Alecsis:
This is my art of a server with hack club brand running Ubuntu desktop , I am a little rusty with coding and proud of my work. It features 3 modes upon running, one being the classic American Megatrends bios, the Ubuntu Desktop Background, and the last being a crash/warning screen with smoke coming out of the server.
image


leomcelroy:
Thanks for the submission! Do you think you could get the corners of the server to line up? Also try to cover up the edges that are obstructed.

leomcelroy:
I like the different screens a lot! Particularly the smoke with the warning screen.

Alecsis:
Thank you for the review! I updated the code to make it more tidy but I had problems with the top left line not being level. Everything else I centered and corrected!

image (1)

Alecsis:
I updated the snapshot directory to reflect the changes.

Alecsis:
Actually, I was able to correct the code making it nicer looking!

image

maxwofford (changes requested):
Hey @Alecsis– this is a great start, but doesn't quite fit the generative requirement for a blot submission! You can have the if statement for different monitor screens (personally i think you should check out `switch/case` statements), but maybe you can also have the monitor change sizes and aspect ratios!

Alecsis:
Hi @maxwofford I've added a wide monitor and a vertical monitor to the 3 different screens, with some text and other small detail changes!

maxwofford:
I'm out and about right now so it'll be a bit of time before I can review this new pass. Good work on making those changes though. That said, it still seems like most of your generative changes are just if statements. Part of the magic of algorithmic art is the change created by random variables, not just Boolean true/false statements. Keep playing with it and adding things that aren't picking between premade configurations & I'll check it out when I can (tomorrow around this time at latest)

BrightTheBackpack (changes requested):
Hey there! 
It looks like the randomness of the monitor causes it too not line up with the stand and screen.

Alecsis:
Ah, how would I fix this?

BrightTheBackpack:
You could either reduce the randomness of the screen a little, or you can make the screen contents move with the monitor

maxwofford:
Hey @Alecsis, how's it going with your art?

Alecsis:
Hi Max! I'm still working on it but I've been from home but I'll try to work on it when I'm available!

Alecsis:
@maxwofford I am back now and ive updated the limits on the random length generator so that the screen isn't broken!

qcoral (changes requested):
This is great, but I think you could add more generative elements to this art piece; it seems to cycle through the same few designs in the same few positions. Could you try editing it to be more generative?

Dongathan-Jong:
Closing this PR due to inactivity! Please create a new PR when you update your art!

PR #633, opened 2024-07-05
funDonutShapes-anvay

anvayajmera:
My artwork is designed to create different donut shapes and sizes. It is focused on the sprinkles (rectangles + triangles), the size of the donut, and the number of swirls on it!

I am a relatively new coder and I spent a lot of my time first learning how to use the Blot!

donut#1


anvayajmera:
Thank you

leomcelroy:
You have copied a significant amount of code from another piece without attribution. Such action can result in being banned from all future Hack Club "You Ship, We Ships". I'll give you just a warning for now.

PR #636, opened 2024-07-05
plants-nash

nash-pillai:
I wanted to create something that looked organic like in the examples that were shown. I did some research and came across Lindenmayer systems. After learning about how they work, I wrote my own implementation here. Once I added the plants, the top of the frame seemed empty, so I added some random stars. The stars are positioned randomly and each point is randomly offset to give it an imprecise hand-drawn look. In addition to the random seed, you can customize the angle range that the branch angle is randomly selected from and the number of each kind of star. Also, you can change the rules to make an entirely different type of plant, but I found that this worked best. Thanks!

image


leomcelroy:
Nice, I think the plants are the real star. How could you make it look more like a field?

PR #639, opened 2024-07-06
mapmaker-Lovidu

Raptor-Coder:
This is my second attempt a coding with blot. After some practice I (and help from fellow hackclub members) I was able to create this, It is a map maker. Sometimes it looks like weather maps and sometimes like countries so it is up to interpretation. I don't have much experience coding and only started this year.


leomcelroy:
Nice! Congrats you've earned a Blot. You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #642, opened 2024-07-07
PokeballShaking-Sriven

srivenc:
The artwork is pokeball in the process of catching a pokemon. To depict this the pokeball can be seen rotated to depict shaking as well as shake lines as seen in the games. Finally, if a certain condition is met, stars show up above the pokeball to depict that the Pokemon has been caught, similar to the games. Coding this was quite satisfying because of the challenge of the shake lines. It was interesting to see how such a small detail can cause such a major challenge, but the catmullRom function proved to be a little bit tricky in combination with the other toolkit functions. Overall, it was extremely gratifying to see the finished product and am curious to see how far I can take Blot.

caught


leomcelroy:
I like this, makes me nostalgic. You misnamed your snapshots folder with a capital S. I fixed it for you.

PR #646, opened 2024-07-08
PolarPlotter.js-StefanLighezan

stefanlighezan:
PolarPlotterSnapshot

My artwork is a polar equation grapher. I made it because when I was learning Precalculus in the 7th grade, I was always stuck trying to see if I graphed a polar equation correctly. My experience coding this is was very pleasant. Within an hour, I had a working circle grapher (r = x, where x is an integer), and from then I added limacon, rose, and spiral support. I will use my Blot to now help my brother who is going to Precalculus in two years not face the same trouble of scouring the internet for a polar graph plotter. The image above has no background when I downloaded it from Blot, apologies for that!


leomcelroy:
I like the idea of drawing something with polar coordinates but right now this is a bit of a "tech demo" try to make an artistic piece with it.

I'm going to close this for house keeping but you can resubmit again later.

leomcelroy:
Also your submission structure (filenames and folders) are incorrect. Pay close attention to the submission guide.

PR #650, opened 2024-07-09
flower-gwentigges

gtigg:
I made a flower generating program. The petal shape is random, and the number of petals is random. I just messed with the numbers for a long time to look right.  The middle is always 20 copies though. And it was really hard to get the size of the middle right.  The stem changes how curvy it is.

flower1


leomcelroy:
This is a simple piece and similar to other flower generators we have but the results are quite nice.

Good work.

PR #652, opened 2024-07-09
LeaflessTree-KarolinaWisniewski

DrasticLemons:
My artwork is of a tree with a unique pattern of branches. The branches are bare and without leaves, while the trunk has some texture to it. My experience coding was very enjoyable, as I had only ever used JavaScript a little bit a few times prior, and so I was able to learn some new things while making this project. 
snapshot2


leomcelroy:
Nice job elevating the simple fractal by giving the tree a trunk.

PR #653, opened 2024-07-09
Walkers-ChristopherBody

cookiemonsternz:
This is a piece of somewhat abstract art. I wanted to express how a landscape can be defined by those who explore it. It creates a configurable amount of 'Walkers', who traverse a distance field (simple noise here) by sampling neighbouring points and depending on configuration, try to either minimize or maximize the sdf value.
I have been coding for several years and have intermediate experience coding JavaScript, python, C and C++.
snapshot0


leomcelroy:
A bit more abstract than what we are encouraging people to do but it looks like some thought went into the implementation. 

Congrats you've earned your Blot! You can order it [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw).

PR #659, opened 2024-07-09
sheet_music-nicholasz2510

nicholasz2510:
My artwork shows a treble clef with 1 to 3 eight notes, with each of the shapes connected as a single line. The positions of the notes and the shapes of the curves are randomized. 
sheet_music


leomcelroy:
Interesting piece. I think it works really well.

PR #660, opened 2024-07-10
SpiderWebStar-Nathan

agentbobjelly9228:
I had a lot of fun coding this piece. I learned how do draw this cool spiderweb star using math when I was in the 3rd grade. It's the only thing I'm actually able to draw well by hand. Now that I learned how to use blot, it's amazing to see how with programming, even I can become an artist.


leomcelroy:
This piece is quite simple but the effect is appealing. While perhaps a bit brief I think it's a good example for the gallery.

PR #670, opened 2024-07-11
UnionedPolygons-sosenteam

Sosenteam:
I'm relatively new to art, coding, and trig. I figured out how the library worked in a few hours but spent a while figuring out how to draw a circle with only points, as well as getting the unioning to work. I made a PR request yesterday, but took it down because I wasn't happy with the result and wanted to make it more artistically pleasing. (Previous PR Request was named "InterlappingCircles-sosenteam"). This artwork is a collection of points with spaced rings around them, and the rings get merged with rings of the same level (distance from point). While unintentional, I found that if you lower the circle resolution really low, you could get cool designs based on other polygons (like in image2).
image
image2
image3


Sosenteam:
Saving the second image to png made it a little less clear, so heres a similar version w/ less clutter
image(2)

leomcelroy:
Thanks for the submission. I like this idea a lot and think it looks pretty cool but the code is still in the realm of what a LLM could spit out. How could you make it a bit more sophisticated and add more human touch?

maxwofford (changes requested):
OoooooOoh, this looks trippy. Love the start @Sosenteam, but as @leomcelroy mentioned I think you can go even further with this!

Sosenteam:
Updated! I'm proud of this one, if there is anything else that you want me to add, I'm happy to keep working on it!

maxwofford (changes requested):
Ooooh, love the color pallet you've got going. & the offset stuff you did in [`6ba3e3a` (#670)](https://github.com/hackclub/blot/pull/670/commits/6ba3e3a184b2fe558bf33ce1c1475ae6d97bd695) looks great!

That said, keep at it & check out https://turtletoy.net/ for inspiration– you're like 75% there. I think you can push yourself even more on this.

Sosenteam:
Updated again, I might try to make it look a little more like an optical illusion, but i've gotten some pretty convincing 3D effects by accident. 
Pictures:
cover
image (3)
image (4)
image (5)
image (6)
image (7)
image (1)

maxwofford (changes requested):
Nice work on updating this art! Heads up though, without color it'll look like this:
Screenshot 2024-07-14 at 03 46 45

See the black spots? that'll probably tear through the page because the pen is going over the same spot multiple times:
Screenshot 2024-07-14 at 03 46 50
I think this might be a bug!

Sosenteam:
> that'll probably tear through the page because the pen is going over the same spot multiple times:
Yeah that center part getting repeated was totally a bug 🪲, should be fixed now . But when I'm 
actually making a print i plan on using a little larger gaps between rings so it'll print better. Like this:
image(5)

maxwofford (changes requested):
Nicely done! Please update the snapshots so I can merge this (I don't see color or triangles/hexagons in the generated art anymore)

Sosenteam:
OK that should be updated!
> (I don't see color or triangles/hexagons in the generated art anymore)
I made some with Dodecagons / Hexagons becuase that is still an editable parameter that i think works really well, but i got rid of the crazy color ones :)

maxwofford (approved):
Nice work on this! Go ahead and [order your blot](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw)

PR #673, opened 2024-07-11
4d_polygons-aryan_sehrawat

123Aryan456:
4D Polygons: A Balancing Act of Complexity and Clarity
The creation of the "4D Polygons" was a balancing act: how much of the impression of the higher dimensions should it give without the visual representation itself turning incomprehensible. It was all a matter of evoking this feeling of intricate structures existing beyond common perception but using nothing but two dimensions and the viewer's own mind.
For me, selected points scattered on the canvas became my building blocks, representing vertices of this imagined 4D space. It was with the lines that the real magic was. Based on proximity, I could also join points to simulate edges of unseen polygons. Now, finding the right density became crucial; too many lines create a tangled mess, too few and the sense of complexity vanished.
The fading line weight is the most obvious tool that misleads a viewer that a great depth lies beyond the flat canvas. Here, the lines that connect more distant points are much more subtly faded away, creating something like visual hierarchy, thereby carrying the wandering eye of the viewer in tracing paths and perceiving layers that wouldn't have existed if not for this carefully calibrated fade.
Another level of constraint was added because I decided to keep the artwork compatible with pen plotting. I needed to be aware of the density of lines and overall complexity, ensuring that the piece would still be striking when translated into the precise but limited language of pen on paper.
"4D Polygons" does not pretend to represent a dimension in its totality, for that it can never capture. It gives one a glimpse, a visual whisper of what such a complexity might imply. One is invited, if not by the work itself, to trace it carefully and piece together these hints of other world constructions in order to realize, finally, how enormous what we can conceive but not perceive fully really is.
Screenshot 2024-07-11 at 10 19 25 AM
Screenshot 2024-07-11 at 10 19 09 AM
Screenshot 2024-07-11 at 10 18 50 AM


leomcelroy:
The files were not submitted properly.

PR #678, opened 2024-07-11
treegen-elijah

grandell1234:
This is my first time coding a Blot script, and one of my first experiences using Javascript in more than just a small backend system. It is a tree generator, you can set a seed and it will create a tree-like object to be "sort of random" like this:
```
function random() {
  const x = Math.sin(seed++) * 10000;
  return x - Math.floor(x);
}
```
and then it changes the things according to the seed, so if you have the same seed it will look the same every time:
```
const newLength = length * (0.7 + 0.3 * random());
const angle1 = angle + (Math.PI / 4) * (0.7 + 0.3 * random());
const angle2 = angle - (Math.PI / 4) * (0.7 + 0.3 * random());
```


grandell1234:
tree
Example image

leomcelroy:
Thanks for the submission. This piece is a bit too simple at the moment.

Feel free to resubmit later when you are ready.

PR #682, opened 2024-07-12
CyberTruck-Viv

GeoWizard4645:
I really enjoyed the blot project. It was a great fun to mix art into coding and I hope to make more things on Blot. 
SnapShot1


maxwofford (changes requested):
Hey @GeoWizard4645! This is def cool art, but doesn't quite meet our [submission criteria](https://github.com/hackclub/blot/blob/main/docs/BLOT_REVIEW.md#submission-criteria)– you need something that generates new art with each load and is algorithmic. What if you drew the cybertruck in a city that generates differently everytime?

GeoWizard4645:
Do you mean like a new different background every time, that’s a good idea?

maxwofford:
@GeoWizard4645 that's a great idea! Yeah, change up the background if you want!

PR #685, opened 2024-07-12
stringArt-jamesLian

James-Lian:
Hello! My name is James Lian, and I made this custom string art generator on Blot! 
This is how it works:
1. Input an image array (only black-white values, 0 for black and 255 for white). My image arrays were generated from a custom Python script that I wrote for this Blot work specifically - you can find it here , or in `index.js` where it is commented. Fair warning that the string-art-generator should work better with more **higher-contrast portraits** (the results also look better if you look at it from afar). 
2. You can play around with the `size` of the drawing (e.g. 300x300, 100x100), the `scale` (up to 0.99 - this is how much of the drawing space the frame of the string art takes up), the number of `pins` (where each of the lines/strings are anchored), the `numLines` variable (how many lines/strings), and finally `stringArtContrast` (between 1-255 - if the number is too high, the artwork will be muddled with many lines. If the number is too low, then there won't be enough lines to make out features).

To explain in simple terms how the algorithm works, I start from a pin and find all possible paths to all other pins. Then, I calculate the darkness of the pixels where each path runs over - the 'darkest' path is drawn and then the pixels where the path passes over is lightened to prevent too many lines from passing through the same area (this behaviour is controlled/limited by `stringArtContrast`. I loved the Blot editor - I think it's so cool! At the same time, I spent 10+ miserable hours debugging only 100 lines of code - my experience working with the Blot editor was a nightmare 😅 although I think it was just my fault for choosing such a topic. 

At the end of the day, though, I'm super proud of my work! There was a time where I thought about giving up, but I'm so happy that I made it through in the end :) And personally, I think the art that I generated was super cool! I'm so glad I thought about adding `stringArtContrast` - before that, I was 'erasing' all the pixels that the lines drew over with white pixels. That was a mistake, as the algorithm was less inclined to draw lines in the darker areas and more inclined to move onto less darker areas of the image where there shouldn't be as many lines - this is where the result will end up as a muddled mess of lines everywhere - even in the lighter parts where there shouldn't be lines at all! The `stringArtContrast` variable allows me to control the amount of which I lighten the pixels - it sounds simple, but in the midst of debugging, when I finally thought to add this in, I was practically by a miracle. 

Thank you in advance for reviewing my program!

(P.S. below is an example image of a string art generated of Mona Lisa. I also included 2 pieces of string art of Mona Lisa within the snapshot folder - one with a relatively high resolution (more contrast + lines), and one with a relatively low resolution (less lines. Fair warning, the pictures exported from Blot look quite messy - I think they look better when rendered in the Blot editor, or maybe if you zoom out. The face of Mona Lisa definitely registers better if you look from afar. I hope you have fun playing around with my artwork!)
mona-lisa


maxwofford (changes requested):
This is soooooo cool! I really love that you figured out how to do this & added controls to help prevent too much overlap. The blot will def chew through paper that's not thick enough so I'd love to play with that setting once this is merged in.

It looks like your code is in that python repo you linked. Just like my feedback for [this awesome maze art](https://github.com/hackclub/blot/pull/621#pullrequestreview-2176313127), you'll have to convert your stuff into js so other people can run this in the gallery and experience it without downloading a seperate codebase!

James-Lian:
Thanks for the feedback! When you say converting the Python script into JS, do you mean linking an external webpage/javascript program? Because I would not be able to imagine doing so in the Blot editor 😅

maxwofford:
@James-Lian you can also do this in your own editor then copy it to the blot editor! I just need all the code in 1 place! We might show the gallery of projects off to people at events & they need to be able to open up a gallery piece, click "run" and see a new thing generated. If there's only the build output in the blot file, then they can't mess with it or try getting it to generate other things!

James-Lian:
@maxwofford Sorry about this but I'm still a little confused.
Would it be acceptable for me to insert the URL to a website that performs the function of the Python script?  
I am asking because currently the Python script is performing 2 functions: 1) allowing the user to upload their own custom image, and 2) converting the user's image to an array of values.  
I've tried and I haven't found a way to allow user's to upload their own images within the Blot editor, as I don't think the Blot editor runs any JavaScript that does anything other than drawing-related functions (for example, I tried and I couldn't open a file dialog, although correct me if I'm wrong).  
And pasting external JavaScript into the Blot editor? Uhh... I mean it's definitely possible (if a little messy), and as mentioned above the code wouldn't work.

maxwofford:
@James-Lian the intent is that your blot submission's core tech (the formula to figure out string position and ideal overlap) is in JS so other people can check it out for themselves. I'm cool if you figure out another way to make it work. I just tested and it looks like fetch requests work fine in blot, so you could do something like this:

```js
// change this to your own art!
const artURL = 'https://cloud-flsp2xrdt-hack-club-bot.vercel.app/01200px-mona_lisa_2c_by_leonardo_da_vinci_2c_from_c2rmf_retouched.jpg'

fetch(artURL).then(art => art.blob()).then(data => /* run your drawing code */)
```

James-Lian:
@maxwofford Thanks for the advice last time! It pointed me in the right direction (eventually)...

Unfortunately, fetching the image from a URL directly within the Blot editor would not work because, for some reason, attempting to execute the `drawLines()` method after the await keyword would result in 0 lines being drawn (mysteriously). The await keyword is needed because `fetch()` is inherently asynchronous, and I tried to circumvent this with `XMLHttpRequest` instead, but unfortunately, to access the image data from a URL, I need to use the `createImageBitmap()` method, which is _also_ inherently asynchronous, and there is no alternative method that I could find for this one... 

So, in the end, I decided to create a [serverless function](https://github.com/James-Lian/string-art-image-service) with Vercel, retrieving the necessary data with `XMLHttpRequest`. I hope that is acceptable! Please let me know if you have any concerns.

maxwofford (changes requested):
Hmmmm, I'm getting an error when running this: `"IndexSizeError: The source width is 0."`

Happy to help with dealing with async stuff though! You probably had your drawing code run outside the async `.then`. Here are 2 examples:

```js
// this doesn't work b/c async (you mentioned):

const data = fetch('url')
const lines = processLines(data)
drawLines(lines)
```

This would work though!
```js
fetch('url').then(data => {
	const lines = processLines(data)
	drawLines(lines)
})
```

Happy to help with more async code if you can't get it working!

James-Lian:
Could you let me know more about the error you encountered? I copy/pasted my code into the editor again, and it seemed to be running fine... 

Were you perhaps using a different image URL? If so, could you provide it?

James-Lian:
@maxwofford Thanks for the suggestion, btw! To clarify, I was indeed running my drawing code outside the async `.then`, but that was only because I was using the `await` keyword as well.

But curiously, even when using the `.then` keyword, the same problem arises.
Here's some sample code:
```
// draw a line from top-left to bottom-right
drawLines([[[0, 300], [300, 0]]]); 
fetch(imageURL).then(art => art.blob()).then(data => {
  console.log('happened');
  // draw a line from bottom-left to top-right
  drawLines([[[0, 0], [300, 300]]]);
});
```
Assuming it is not just a problem on my end, if you run the above code, the console will log 'happened', but only the first line will be drawn. For some reason, the second line, from the bottom-left to the top-right, will not be drawn 🤷‍♂️ 

In case you were curious, below is my previous code using `fetch()` and `await`. The same problems are present here 😞 
https://github.com/James-Lian/blot-string-art-generator/blob/main/project%20(6).js

maxwofford:
@BrightTheBackpack can you help with this submission either here or in slack?

maxwofford:
Just posting this as an update– we have a discussion right now on the team about how to handle async code. Sorry for the wait!

qcoral (approved):
This is insanely cool - I'm merging this for now as I think it deserves it, but I'll reach out on slack to see how we could possibly resolve the async code stuff. Congratulations on your Blot! You can order one [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw)

James-Lian:
Thanks so much!

PR #707, opened 2024-07-17
Birb-AyaanGrover

ayaangrover:


ayaangrover:
Hey Max, 

Thank you for the feedback. I have taken your feedback into consideration and adjusted the code accordingly. Please do let me know what you think of the updated code once you have reviewed it.

Sincerely,

Ayaan

BrightTheBackpack (changes requested):
nice improvements!
 How about making it so there's a random amount of birds on the branch

maxwofford (changes requested):
Nice work! It's close, but there's still a line that's a little off here– I think the line for the tail is travelling with the pen down instead of up? Not sure, but it looks like this on my screen:
Screenshot 2024-08-02 at 11 47 07

Other then that, you should make it more generative! I like how the tail moves each time, but the rest of the bird should change shapes/sizes/patterns too!

ayaangrover:
After almost two weeks of work, I have updated the code with the biggest update so far. Now, there are a random number of birds on each electrical line. There are also random number of electrical lines, between 1 and 3. Also, there is a random sized sun in the top left with a random number of rays and a random ray length. The birds leg length is random and the bird direction is random. As a bonus, in the console the code logs the unique ID assigned to the specific generation.

ayaangrover:
I have updated the snapshots to display the most recent code update.

maxwofford (approved):
This looks so great now! Nice work. You can [order your kit here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw)

PR #710, opened 2024-07-18
cuteOneLineFlower-emma

emma-x1:
Generates a three-petal colored flower. Uses random seed value to move the position of the flower, stem, and leaves, as well as change the shape of the petals slightly. This is my first coding project and I'm a beginner coder :)
line_flower3


BrightTheBackpack (changes requested):
Wow this is a pretty flower!
It think it would be cool to see a flower field with multiple random flowers.
It would also be cool if some flowers had 1 leaf, and some had 2

BrightTheBackpack (approved):
nice! Max will be back soon to do a final review

maxwofford (changes requested):
This looks great but I think you should step it up a little bit more! Make the petals change their size individually, give the leafs veins, that sort of thing!

qcoral:
Seeing as this hasn't been updated in over a month, I'm going to close the PR for now. You're free to make a new one in the future with the same artwork if you'd like! Reach out to me @ alexren on the slack if you have any questions!

PR #717, opened 2024-07-18
Significantly improve error messages

jeremy46231:
This PR improves error messages significantly:
Line 27: bt.rotate expects argument 1 to be an array of polylines, but got a single polyline

Before this change:
![Line 0: undefined is not a function (near '...[o,u]...')](https://github.com/user-attachments/assets/8bfef329-4530-485f-b0c5-6a96f6337746)

This adds assertations to all of the built-in functions in Blot to throw meaningful errors when the wrong arguments are passed. It also upgrades the stack trace parsing logic, using V8's object representation and a regex built based on the SpiderMonkey and JavaScriptCore source code. Safari (JavaScriptCore) does not keep track of line numbers within eval'd code, so a hack is used to construct a data URL and import it when it is detected to be nessesary.

The previous version of Blot would give unhelpful errors like "t is not defined" on line 0, and line numbers only ever worked on Chrome in some cases. Now, it will give helpful errors explaining what went wrong, pointing to the correct position in the code on any browser.

There is a known issue where, on Safari, the stack does not follow nesting, so if the user defines a function, the error will appear where the function is called instead of within the function, but
there is no way to fix this.

### DIFF
--- a/src/drawingToolkit/Turtle.js
+++ b/src/drawingToolkit/Turtle.js
@@ -1,10 +1,13 @@
 import { translatePt, rotatePt } from "./affineTransformations.js";
+import { assertArgs } from './assert.js'
 
 const copy = obj => JSON.parse(JSON.stringify(obj))
 
 export class Turtle {
 
   constructor() {
+    assertArgs(arguments, [], 'new bt.Turtle')
+
     this.drawing = true
     this.position = [ 0, 0 ]
     this.angle = 0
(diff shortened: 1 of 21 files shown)


maxwofford:
This is really awesome @jeremy46231! I'm traveling for a week so I can't merge this in, but I'd love to test this out once I'm back in town

maxwofford:
Nice work on this @jeremy46231! I'm going to merge this in and see what the community thinks.

PR #731, opened 2024-07-21
Pepe Mosaic - NotIshi28

NotIshi28:

- [x] Optional, if you remixed this from something else, mention it here:


BrightTheBackpack (changes requested):
Hi there! Nice piece, i really like the mosaic effect. 

- Do you think you could make the mosaic generative, so that every time its ran the pieces look a little different
- Could you also move your snapshots into a folder called snapshots

NotIshi28:
made the changes

PR #745, opened 2024-07-26
Add features to SVG import

whatwareweb:
This PR adds new features to the blot web interface! Now, when you import an SVG, you will have the option to automatically scale it to a specific size - no more randomly guessing scaling and translation values!
Also supports adding a small bit of padding if your blot has issues drawing around the edge.

image


maxwofford (changes requested):
Looks like this PR accidentally added a bunch of whitespace floating around. Could you fix that so I can merge @whatwareweb ?

whatwareweb:
@maxwofford Whitespace removed.

maxwofford:
Nice work! Quick thing for improvement:
- it's a little funny it's possible for an invalid input. Can you just autoscale a non-filled height/width?
Screenshot 2024-07-29 at 00 09 23
- actually adding the polylines is kinda hard to copy from the box. maybe a button?

PR #750, opened 2024-07-27
FlowingCurrents-KfirElyahu

BoghosDavtyan:
Hi, I'm Kfir! This is my first time making art with JavaScript. I called it "Flowing Currents" because it was inspired by how wind or water moves. It was tricky to make the smooth, wavy lines fit inside the canvas, but I'm happy with how it turned out. The lines were made using Perlin noise, and they look like abstract pictures of natural stuff or even space energy. It was really cool to see how changing things like the noise scale and curve strength could completely change the picture. This project got me really interested in making more art with code, and I want to try more complicated designs in the future!

- [X] I have read the steps to [getting a blot](https://github.com/hackclub/blot/blob/main/docs/GET_A_BLOT.md)
- [x] I am submitting art that...
  - [X] is algorithmically generated (will change each time the program is run)
  - [X] is drawable on a blot (fits in the work area & doesn't overlap too much)
  - [X] is original (not copied from somewhere else)
  - [X] doesn't call `Math.random()` (See the documentation on [randomness](https://github.com/hackclub/blot/blob/main/public/TOOLKIT.md#randomness))

- [ ] Optional, if you used a tutorial or based your art on something else, please include the link here:
- [ ] Optional, if you remixed this from something else, mention it here:

PR #754, opened 2024-07-29
Revert "Significantly improve error messages"

maxwofford:
Reverts hackclub/blot#717


maxwofford:
@jeremy46231 the highlighting works great, but your assertions aren't correct– they were marking a bunch of false positives on art that generated fine. Try out https://github.com/hackclub/blot/pull/750 for an example!

maxwofford:
lmk when that's updated because I love the new warnings this gives!

PR #756, opened 2024-07-29
BlotChess - Random Chess Puzzles now with Blot!

blazecoding2009:
**REVIEWERS PLEASE WAIT 15s-30s BETWEEN EACH GENERATION FOR THE API TO REFRESH**


maxwofford (changes requested):
Screenshot 2024-07-29 at 09 42 18


This looks amazing! Nice work on it. Just a couple things to change:

- some of the black pieces have a lot of overlapping lines (image above) which may cause the blot to tear the page
- some of the pieces are hard to tell apart (R & B), maybe you can update the art a little more to make it more aesthetic and clear?

blazecoding2009:
Hey Max! Thanks for reviewing this
The Black pieces are supposed to do that to create that illusion of a dark piece. I wasn't sure how to go about this. I'll reduce the time it goes over it and increase the spacing a tiny bit more (from 10 to 4)

As for the other one, I can see where you're coming from, I'll go ahead and make them a bit different

blazecoding2009:
Ah I see, I'll try to do something with that. While I do that, could it be possible to order my blot?

PR #764, opened 2024-07-30
BlotTacToe-spageektti

spageektti:

I didn't use any code from another project, but the idea was inspired by the Blotteships project.


qcoral (changes requested):
The technicality of this project is really interesting - could you maybe add a short explanation/description about what the program does and how you made it? Other than that, I think you can turn this into something even better; keep going at it!

qcoral:
Seeing as this hasn't been updated in 2 weeks, I'm going to close the PR for now. You're free to make a new one in the future with the same artwork if you'd like!

PR #765, opened 2024-07-30
ParametricRandomF1CircuitGenerator-AbhinavMalladi

ShutBaton:

I am relatively new to JS, only having used it for Sprig and a little trivia game on my website in the past (almost a year ago). I have dabbled in other languages. The program may not be very long but took me a long time to figure out and tune to produce good results. Ordering each control point for the curve by its angle from the midpoint on the page was a revelation for me that only appeared halfway through the process of making it for me. Beforehand (and sometimes afterwards) I rewrote the whole thing to be more efficient or produce something that looks better. I am very proud of the result it makes - even if it took me FAR too long to actually code it up. :)

The below track is fully generated by the program, sized for an A4 page. To be more realisitic (and possible with a pen) this can be sized down parametrically as the program will adapt the drawing width, height and pen size accordingly.

Track


ShutBaton:
Made a better version, so closing this PR to open a new one with V2!

PR #775, opened 2024-08-02
ParametricRandomF1CircuitGenerator-AbhinavMalladi

ShutBaton:

Uses the Text Engine in Scott C's and in geschmit's submissions. The repo for the engine is [here](https://github.com/geschmit/blotfont).

I have been coding for a little while now, with some experience in Python and with C++.

I am relatively new to JS, only having used it to make a small web trivia game before, or to do a few API calls for a group project. The syntax similarities to C++ sort of threw me off a bit, as I would often accidentally switch into using C++ syntax during the art's development, like suddenly typing `pair<int, int> coords;` instead of just `const coords;`. I definitely spent at least 8 hours on this project as a minimum! It took me far too long to get the F1 Logo .svg file into the art, and I ended up learning alot about how svgs work in that process!

The art is entirely parametric and so should scale to any size of page appropriately. This is version 2, as I was not happy with the generation and look of my previous version, which had a thick, filled in track (also not really possible on blot) and the lack of much branding. I am very happy with this version of the track generator, even if it gave me hours work as I figured out how to make it generate the track randomly, and also make use of the toolkit as much as possible. I've explained it all in comments within the code. :)

Edit: Features Added -

- Higher Res F1 Logo with registered trademark symbol added
- Page borders added
- Start Line added
- Customisable or Random (controlled by constants at the top of the program) race text.
- This text fully scales to page size and can be customised in many ways.
- Purpose of the project now different: "to make a fantasy F1 track poster to display"

By the way, the picture below is set to be at A4 Size, can easily scale to a 100mm by 100mm page and look nice. I've set the race info to be something different to show customisation.

image

The image below this shows variables that can be edited to change how it generates:

image


ShutBaton:
By the way, by default it is set to draw on A4, but i realise blot may not have that capability (I don't know its dimensions), so I ensured it is guaranteed to look nice on a 100mm x 100mm canvas!

maxwofford (changes requested):
This is a cool start, but I think you can take it further!

What if you show the starting/ending line, or have the onramp for starting the race. OOOoh, or label different areas and annotate them:
Screenshot 2024-08-02 at 14 45 36


Maybe even come up with a name for each of the generations that makes a random name for the prix:

Screenshot 2024-08-02 at 14 47 01

ShutBaton:
Because of the way the generation works (picking random points and then creating a curve around them), I'm not sure how I would create straights or label corners without remaking the entire thing to generate in a whole new way. I will try and upgrade this further, adding details like a track start and more like you mentioned, however maybe not exactly by adding information like corner labelling or DRS zone labelling. Thanks for the advice!

Image showing the actual random line overlayed on the curve drawn: 
image

ShutBaton:
@maxwofford I have now upgraded it as much as I can! I have added the start line and added text as you have said, but also made some more improvements highlighted in the now edited comment above. I am very happy with the result :)

In total this project has taken approx. 14-15 hours so far, most of it being put into doing the maths for the scaling, learning how to use the text engine, how an .svg works under the hood, and getting better at JS overal.

ShutBaton:
@maxwofford Is there anything else I need to change? I've done everything you mentioned + a bit more, except for the zone labelling as that would require rewriting how it generates in a complex way I'm not sure how to implement yet. Thanks!

qcoral (approved):
Nice job on making the adjustments! This is a great piece of clean, generative art. Congratulations, you can order your blot [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw)

PR #778, opened 2024-08-02
CanadaFlag-Shaurya

madebyshaurya:
  - [ ] is algorithmically generated (will change each time the program is run) 
  **PS: It's the Canadian flag and it is a constant , so I didn't see a way I could make something change or else it would not look like the flag. I hope this is ok.**

- [x] Optional, if you used a tutorial or based your art on something else, please include the link here: Canada Flag

I live in Canada so I drew the Canadian Flag. I think this is cool since it is almost impossible to draw a maple leaf free hand perfectly so this was a great experience. This might not be that accurate since I eyeballed the angles and the length but it looks pretty much like the flag. First time using blot and I was really amazed by the features Hack Club team has put in. I mainly loved the turtle.

Here's the final product :)


maxwofford (changes requested):
This looks awesome, but unfortunately this won't get a blot machined unless you make something generative!

maxwofford:
What if you made a scene with a flag in it?
Screenshot 2024-08-09 at 15 48 41
Screenshot 2024-08-09 at 15 48 52

Dongathan-Jong:

PR #780, opened 2024-08-02
Sunflower-Parv

goyalparvv:
This is my art piece called "Sunflower." It shows a sunflower that changes every time you run the code. The design is inspired by how real sunflowers are unique and different from each other.

I enjoyed creating this piece because it taught me new ways to code. It was tricky to get the petals just right, but after some work, I was happy with how it turned out. I’m excited about how it looks.

Overall experince is great!


maxwofford (changes requested):
Hey @goyalparvv! Please double check the requirements on generating art! You're using Math.random() here even though you checked off that box
Screenshot 2024-08-06 at 16 55 24

Also the art isn't very generative. I'll leave this PR open so you can make edits in here!

PR #785, opened 2024-08-04
spiro_grad-subarctic

sub-arctic:
snapshot_3
*Note: the exported and attached images seem to have a thicker line width than what is visible in the editor, however this isn't a terrible thing*


qcoral (changes requested):
I think this is really pretty, but it's very reliant on color to look good. On an actual Blot, it'd look more like this: 
image

Could you try and make it work without color? I think the patterns could wind up being really cool

sub-arctic:
@qcoral - I've just set the output to a solid color. Let me know if something more detailed or unique is required, and I will be happy to provide :)

qcoral:
This art piece is currently generated in a really straightforward algorithmic manner. Can you change it so that it has more creative intent?

sub-arctic:
I've created a new (better) Blot project, so I will close this one since the new project is a successor and significantly different from the current one.

PR #795, opened 2024-08-07
Darren-Jacob

Deadskullcandy:
For this project, I have developed a piece of generative art that transforms any given image into a blot code representation. The artwork uses algorithmic processes to convert pixel data into drawable lines suitable for the Blot robotic drawing machine. This approach ensures that each rendering is unique, capturing the essence of the original image while adding a distinct computational aesthetic.
Screenshot 2024-08-06 at 20 25 50


qcoral (changes requested):
Hey! I love the technicality of the project, but a couple things:
1. Where is the code for the image converter? Can you include it in the PR?
2. Linear image converters are unfortunately not considered generative enough for Blot. Maybe you could generate an alternative image based on image input?

Dongathan-Jong:

PR #796, opened 2024-08-07
zoyrangutan-zoya

zreese:
"I made an orangutan my favorite animal. It was hard to make it look like a forest so we made lots of trees and then green triangles which look like leaves. There are three different types of orangutans and they're just as smart as humans but there are not many left in the world. I'm ten years old and haven't made art like this before but it was fun and Papa helped."

zoyrangutan2


Dongathan-Jong:

PR #802, opened 2024-08-09
RecursiveBlossomBranch-DevJoshi

devjoshi0:

- [x] Optional, if you remixed this from something else, mention it here:
I used Hayden Kong's "Artistic, Unique Tree" blot code as a reference to mine, you will be able to see it on the same name for the functions and variables, but I did not copy paste cause I wanted to make something unique, Instead of rendering a full plant I wanted to create a branch that went from left to right with blossoms.

This project uses recursive logic to generate a tree branch going from the left side to the right side, it first generates a random seed which is used to make every tree generated unique, There are parameters which you can edit to make the tree look different depending on how you want it, for example you can adjust how much blossoms are created using the "blossom density" variable or by adjusting the trunk width. I liked the look of blossom trees so I wanted to make my project unique by adding in star shapes to mimic it. 

I started coding when I was 13, but I haven't gone fully serious until my 11th or 12th year. I love working with physics and making games using the Godot Engine, I wouldn't say I have intensive knowledge but i'm also not a beginner, somewhere in the middle. :D
Screenshot-3


qcoral (changes requested):
Hi! This is a really great start! I think it looks great, but there's a couple things you could improve:
1. Could you center the branch so that the majourity stays in the center?
2. It currently feels a little unpolished, I think you can clean up the drawing a bit!
Comment when you update your PR and i'll come back around!

Dongathan-Jong:
Just an update, I will be closing this PR. Please create a new PR once you have updated your art! We are doing this due to new management of the PR system.

PR #817, opened 2024-08-11
AirRoutes-Desert

GameDesert:
# AirRoutes

snapshot1

I found the visual of flight paths over a map quite interesting, almost like a spider web across the Earth, so I wrote a script that converts a list of routes into line segments that can be plotted with Blot (`https://github.com/GameDesert/BlotAirRoutes/blob/main/airports.js`). I turned this into an API and called it here in `index.js` to allow the user to draw any shape or pattern they see fit, using purely a list of airports.

## How It Works:

- The user provides a list of routes, providing the IATA Codes (format: XYZ) for the start and end
- Each airport is looked up in a table (provided by `openflights.org`) to find its global coordinates
- A great circle formula is used to calculate the arc taken by a plane on the route (due to map projections, etc) and 3 points are found along the line
- The point global coordinates are converted to relative ones, as a fraction of the X and Y axis of a given bounding box (in this case, Europe)
- These coordinates are sent back to the drawing script, which then draws each arc using the 5 points provided (1 start, 1 end, 3 midpoint) and a Catmull-Rom function

To change the list of flight routes, the format is:
```
...,
{"from":"ABC", "to":"XYZ"},
...
```

Unfortunately, the [script for converting the routes to points](https://github.com/GameDesert/BlotAirRoutes/blob/main/airports.js) had to be run separately as an API due to its reliance on libraries and reading from external CSV files. It is also quite resource-intensive so it would be too slow to run in-browser.

All the files used in the creation of this project can be found here: `https://github.com/GameDesert/BlotAirRoutes/`, but the script in this PR should be able to successfully render the image on its own, given access to the API.

There is a limitation on how many routes can be added before the rendering becomes too slow.

---

  - [x] is drawable on a physical machine (doesn't have lines overlap more than 5 times)


GameDesert:
Yeah it's @Desert on the HackClub Slack.

Dongathan-Jong:
Just an update, I will be closing this PR. Please create a new PR if you would like to update your art! (I'm not sure as to if Alex accepted this PR or not!) We are doing this due to new management of the PR system.

PR #836, opened 2024-08-19
LissajousCurve-Pratyush

Pradevel:
It makes cool art using the Lissajous curve function. (Its randomized every time)

  - [x] doesn't call Math.random() (See the documentation on [randomness](https://github.com/hackclub/blot/blob/main/public/TOOLKIT.md#randomness))

snapshot1


qcoral (changes requested):
I like this! I've never heard of these curves before, so I definitely learned something new with this PR. I think you can take this concept even further through even _crazier_ math (or something else!), so keep working at it! I'll come back soon

Pradevel:
@qcoral  Honestly, I don't know what I changed but seems like we have two curves now making a quite interesting output. I'll update the pics too soon. (updated)
PS: There is a `TILT` variable which allows you to play with the design. You can try sliding the value for some interesting outputs. 

Screenshot 2024-08-19 at 9 02 31 PM

Pradevel:
@qcoral I changed from drawing lines to drawing squares of varying sizes. I updated the pics as well. 
snapshot1

Dongathan-Jong:
Super cool! I think this could use some more work, maybe different shapes? Please create a new PR once you have updated your art! We are doing this due to new management of the PR system.

PR #839, opened 2024-08-20
RandomStarrySky-Ben

SlightlyLessBigJimmy:
My artwork is a drawing of a starry sky with a hill, trees, and stars. Everything is able to be modified via values at the top of the script, and everything is randomly generated. The horizon line will be different every time, and can be modified for different results. The trees and stars are randomly placed around the screen, with the stars staying in the sky and the trees on the ground. 

While working on this piece, I learned a fairly new programming language for me, Javascript. It was also educational to learn about using programming for a new purpose, as I usually use it for game development.

### DIFF
--- a/hardware/motor-control-board/firmware/firmware.ino
+++ b/hardware/motor-control-board/firmware/firmware.ino
@@ -42,9 +42,9 @@ void setup() {
 
   pinMode(enablePin, OUTPUT); // enable pin
 
-  // pinMode(PIN_LED, OUTPUT);
+  pinMode(PIN_LED, OUTPUT);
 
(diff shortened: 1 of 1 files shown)


maxwofford (changes requested):
Nice work on this! Just a couple things:

- Currently it relies on "fill" to cover up parts of the drawing. Keep in mind the pen plotter drawing in pen can't do that, so your art will currently look like this:
Screenshot 2024-08-20 at 15 26 53
- Some generations leak out of the drawing area! Make sure your drawing fits inside the blue lines
- maybe add some more details– this is a good start, but to get into the gallery you'll need a little more detail here

PR #863, opened 2024-08-24
Add files via upload

EuanRipper:
My program takes any image and converts it into something blot can draw. This is done by representing each pixel as a number of dots, where darker pixels get more dots. The dots are randomly distributed within the pixel. I also wrote some python code to process PNGs into an input for this program. This could take a while to draw though. 
image


EuanRipper:
Would it be possible to use the photo in the description for the vote in the competition? I wouldn't expect you to draw it because it took like 8 hours but it shows what blot is capable of with a bit of time! if not the hack club logo should draw in a reasonable time, please be patient though!!
-thanks, euan

blazecoding2009:
Hey @EuanRipper 

Amazing project! Could try try making it a bit faster? You could try decreasing the resolution its doing (instead of 10 dots per, 5 dots etc)

EuanRipper:
Hi! I have made it faster, now it draws less dots but connects some of them to maintain the same density of ink!
image

Dongathan-Jong:
Hey! Super cool idea! Currently we are looking for pieces with 15-20 hours of effort, and it looks like this could use some more! I could recommend drawing out the photo using flow fields, or some other sort of randomly generated pattern. Please create a new PR once you have updated your art! We are doing this due to new management of the PR system.

PR #868, opened 2024-08-26
MoonlitRoad-Ben

SlightlyLessBigJimmy:
My art piece is titled "Moonlit Road" because it is exactly as it says, a picture of a road on a landscape, with a moon and stars. The piece will generate differently each time, with the horizon line changing, the stars being placed and scaled randomly, and the road will follow where it should be according to the horizon line. My favorite part is the moon, however because it will have a randomly decided phase based on 3 different moon types. During the process of making this piece, I learned a new language, Javascript, and a new use for programming, Art! I never thought about making pictures with code before this. 

image


SlightlyLessBigJimmy:
> This is a great start! I think it'd be really cool if you added something to the landscape! It would be **SUPER** cool if the stars had different patterns too! I'm really excited to see what you come up with :D

thanks, will do

PR #869, opened 2024-08-27
AutogeneratedMaze

Babyyoda777:
My code auto generates a random, unique maze each time. Nice and simple code but the result can take a while to solve!

- [*] I have read the steps to [getting a blot](https://github.com/hackclub/blot/blob/main/docs/GET_A_BLOT.md)
- [*] I am submitting art that...
  - [*] is algorithmically generated (will meaningfully change each time the program is run)
  - [*] is your own work that is not copied from somewhere else
  - [*] doesn't call `Math.random()` (See the documentation on [randomness](https://github.com/hackclub/blot/blob/main/public/TOOLKIT.md#randomness))
  - [*] is drawable on Blot itself (doesn't have lines overlap more than 5 times)
    - [*] Doesn't overlap lines more than 5 times
    - [*] Fits in the working area of 125x125mm
    - [*] Doesn't rely on colour to look good

image


Dongathan-Jong:
Really cool idea! You're off to a great start, but I would like too see this with a bit more detail! You could add a start/end point, or make the maze solvable! 

PS. Your canvas size is set to 400, it should be 125 :)

Dongathan-Jong:
I will be closing this PR due to inactivity! Please create a new PR when updated.

PR #878, opened 2024-08-28
Fix SVG imports by implementing the "string" type in assert.js

jeremy46231:
[See thread on Slack](https://hackclub.slack.com/archives/C04GCH8A91D/p1724826463114139?thread_ts=1724826463.114139&cid=C04GCH8A91D)

The bt.svgToPolylines function uses the "string" type in its assertion line, but that type was not implemented. This PR fixes that, fixing SVG import.

### DIFF
--- a/src/drawingToolkit/assert.js
+++ b/src/drawingToolkit/assert.js
@@ -8,6 +8,7 @@ Type strings:
 - function
 - boolean
 - undefined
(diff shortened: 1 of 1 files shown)


qcoral (approved):
Thanks for finding this! Tested locally, merged!

PR #880, opened 2024-08-29
TuxedoCupcake-Olive

EnterpriseGoose:
This is a cupcake that's based on the cupcakes I make in real life every week. I love making a whole bunch of different flavors, so I tried to replicate that by having tons of randomization in terms of what the cake, frosting, and sprinkles would be. There's even a little randomization in what the cupcake wrapper looks like (which is also realistic- not every cupcake is perfect!). I struggled quite a bit trying to get the shape of the cupcake to work, and seeing as this is my first time working with something like this, it took me a bit to understand some of the documentation, and it took a lot of trial and error to get all the shapes right. Especially with the frosting curves, I tried to get both of the curve generation methods to work, but it just kept looking wrong, so I eventually found the turtle and found that it worked better at getting the arcs I was looking for. By the end of working on this, I now feel a lot more comfortable trying to make the program actually make something the way I want it to look, and I'm excited to show this to my club so we can all work on trying to make our own drawings. I'm super excited to be able to get the blot!

Also, as much as I think it does look way cooler with the different colors, there's still some amount of randomization without them (sprinkle placement, base lines), and it still looks really good in black and white. I added a photo to the snapshots that shows what it looks like in black and white (aka what it looked like before I added color).

gallery


blazecoding2009:
Hey @EnterpriseGoose 

Could you try making it a bit more random when for ex, a blot is only doing one color?

EnterpriseGoose:
Hello @blazecoding2009 ! Thanks for looking over my cupcake.

I added a couple more cool things:
- Toggle for switching color on or off so it's easier to see what it'll look like with the blot
- More different toppings: chocolate chips, cashews
- Added collision detection so that toppings don't overlap. This should make it slightly more realistic (I haven't invented quantum sprinkles quite yet), and it'll probably work better on the blot.
- Made it so instead of rotating the toppings by rotating the turt at the start (which meant that the toppings were rotated centered on one edge), the path is rotated at the end instead.

Let me know what you think! (also if you have any more ideas of randomness I could add)

PR #893, opened 2024-09-03
feat: Implement procedural text generation

geschmit:
This commit implements the [Blotfont](https://github.com/geschmit/blotfont) text generation system into the Blot toolkit. I already have written some documentation for the brief toolkit guide and full documentation page, however not all of the art pieces that used the original system have been converted from the original system.


qcoral (approved):
Tested locally, seems to work well for me too. Insane work on this, seriously kudos

PR #897, opened 2024-09-04
sunflower

RealBob64:
This was artwork was inspired by the sunflowers that I grew in my backyard.

- [ Yes] I have read the steps to [getting a blot](https://github.com/hackclub/blot/blob/main/docs/GET_A_BLOT.md)
- [Yes ] I am submitting art that...
  - [Yes ] is algorithmically generated (will meaningfully change each time the program is run)
  - [ Yes] is your own work that is not copied from somewhere else
  - [ Yes] is drawable through the editor(does not use outside tools)
  - [ Yes] doesn't call `Math.random()` (See the documentation on [randomness](https://github.com/hackclub/blot/blob/main/public/TOOLKIT.md#randomness))
  - [Yes ] is drawable on Blot itself (doesn't have lines overlap more than 5 times)
    - [Yes ] Doesn't overlap lines more than 5 times
    - [Yes ] Fits in the working area of 125x125mm
    - [Yes ] Doesn't rely on colour to look good

sunflowers


Dongathan-Jong:
Hey! I love the idea of sunflowers, but these are almost impossible to draw! Since there are so many overlapping lines, it would tear through the paper when trying to draw. Furthermore, your design uses color to show the sunflowers, could you remove it as per the requirement "Doesn't rely on colour to look good" 

I would recommend drawing a 2D sunflower, similar to:  
image  
(excuse the bad drawing it was me)

RealBob64:
> Hey! I love the idea of sunflowers, but these are almost impossible to draw! Since there are so many overlapping lines, it would tear through the paper when trying to draw. Furthermore, your design uses color to show the sunflowers, could you remove it as per the requirement "Doesn't rely on colour to look good"
> 
> I would recommend drawing a 2D sunflower, similar to: image
> 
> (excuse the bad drawing it was me)

Hi! Do you have any suggestions on how I can make it "is algorithmically generated (will meaningfully change each time the program is run)?" Also, if I remove the color and made it so the flowers don't overlap, would that be acceptable?

Dongathan-Jong:
Based on what I have seen, I would recommend a few things:

- Number of pedals changes each time
- Number of seeds change and the orientation of each seed
- The leaves on the stem are randomized with different shapes and sizes
- The height of the flower is changed each time
- And maybe even having more than one flower or a whole garden! 

Just remember, the more effort that is put in, the higher chance it has to be approved! 

If you remove just the colour and made it so the flowers don't overlap, there would still be some problems drawing it, as each flower would be undrawable, from personal experience, if a line is less than 0.6 away from the next line, and it goes over at least 3 times, it would cause some trouble with the paper.

Dongathan-Jong:
Just an update, I will be closing this PR! Please create a new PR once you have updated your art! We are doing this due to new management of the PR system.

PR #901, opened 2024-09-06
LineyPeaks-Jannik

playlogo:
## Check list
    - [x] **Could overlap lines more than 5 times** But only for short distances and it would contribute to the wanted look

None

## A short description
This artwork is my first (and completely self-made, thus very inefficient) implementation of a Gas Simulation, made to look nice.

How it works:
- I generate 1-3 randomly rotated two-dimensional functions to serve as the initial pressure values for my gas particles. (This step is only necessary to create art, not simulate a gas :) )
- I then interpolate the pressure between the points to create a continuous map of pressure. (This step can be used to render the simulation.)
- Finally, I use the map created in the previous step to determine where the gas particles "would like to go". These are displayed by the lines.
- [In a real gas simulation, I would now move the gas particles and repeat!]

Although the code isn't extremely long, I've done a lot of research into fluid (and gas) simulations and the (for me still somewhat magical) math behind it.

Even if this art might not be good enough for a machine, I still want to say thank you for giving me the motivation to finally approach this new topic and start learning about it!

A Image:


Dongathan-Jong:
Super awesome backstory into how you created this! This is pretty close to being merged, but falls just a tad bit below the bar! Do you think you could add something that would really give it a wow factor? Once completed I can forward this to Alex for approval :)

playlogo:
Hey, I think I found the missing wow factor! :)
It's now much cleaner and more organized, but it still contains all the scientific background it had before. 
In fact, it now displays the movement of the gas particles and, when looking at it from farther away, also the pressure of the whole gas! The darker it is, the higher the pressure!

Here's my favorite result:


I've also cleaned up the code and improved the documentation!

Dongathan-Jong:
Great! I will forward this to Alex for approval, and it may take anywhere from a couple days to a couple weeks to receive a response. Please be patient!

qcoral:
This is a really cool demo! I think it would be a lot greater if you gave people options to mess with the initial 2D functions more, but it's still really cool regardless. Congrats! You can get a blot [here](https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw)

PR #903, opened 2024-09-07
TuxedoCupcake-Olive

EnterpriseGoose:
This is a cupcake that's based on the cupcakes I make in real life every week. I love making a whole bunch of different flavors, so I tried to replicate that by having tons of randomization in terms of what the cake, frosting, and toppings would be. There's even a little randomization in what the cupcake wrapper looks like (which is also realistic- not every cupcake is perfect!). I struggled quite a bit trying to get the shape of the cupcake to work, and seeing as this is my first time working with something like this, it took me a bit to understand some of the documentation, and it took a lot of trial and error to get all the shapes right. Especially with the frosting curves, I tried to get both of the curve generation methods to work, but it just kept looking wrong, so I eventually found the turtle and found that it worked better at getting the arcs I was looking for. I then added a few different toppings: sprinkles, chocolate chips, and cashews (which were really hard to get drawn right). The topping choice is randomized, and then they are randomly placed on the cupcake. When it's trying to place the toppings, I originally had it completely random, but I found that it overlapped them a little too much, which I didn't think would draw well on the blot. I then added a collision detection system so they could be a little more separate, and that worked perfectly. By the end of working on this, I now feel a lot more comfortable trying to make the program actually make something the way I want it to look, and I'm excited to show this to my club so we can all work on trying to make our own drawings. I'm super excited to be able to get the blot!

Also, as much as I think it does look way cooler with the different colors, there's still some amount of randomization without them (topping placement, type of topping, base lines), and it still looks really good in black and white. I added a photo to the snapshots that shows what it looks like in black and white as well as color, which can be toggled in the program with a simple boolean.

gallery


Dongathan-Jong:
Super cool idea and great start! Unfortunately, this submission falls short of the requirements (due to simplicity). I would recommend making the icing shape change every time, or having there being a random number of cupcakes inside a case! You can make anything, and I would love to see more :) 

Also, the sprinkles seem to be generating outside of the icing, could you fix that? 
image

Dongathan-Jong:
Just an update, I will be closing this PR! Please create a new PR once you have updated your art! We are doing this due to new management of the PR system.

PR #924, opened 2024-09-27
Rasberry Pi Art

wa1ker38552:
Represents a rasberry pi since I have one at my own workstation. I used python to generate the coordinates needed by taking an image, applying lots of filters to it and detecting outlines/contours on it to output the final list of coordinates which I put in the js file. The section at the bottom of the canvas is meant as a reflection and can be omitted. Any image can be generated as well! 

image


Dongathan-Jong:
Hey! Cool idea, but blot pieces need to be generative! That means that each time you run the code, something about this art piece needs to change. Please look over the [submission guidelines](https://blot.hackclub.com/submitting) carefully and update your PR. Once updated, I will be happy to re-review! (btw your art cannot exceed the size of 125x125, please keep it in that square!)

PR #928, opened 2024-09-29
Robot Generator

wa1ker38552:
Automatically generates an image of a robot using an API. For some reason, drawLines doesn't work unless in the main thread so I wrote a bookmarklet to assist this program. Just create the bookmarklet, paste the code in for the bookmark and for the site and it should automatically run. 

Snapshot (2)
Snapshot (3)
Snapshot (1)


Dongathan-Jong:
Hey! Sick idea, but this piece isn't randomly generated after each run! I understand you would like to use an API, and that is fine, but we are looking for pieces with more complexity made from the user! I would like to see you do more with this art, and would be happy to re-review once updated!

wa1ker38552:
It is randomly generated, the robots are generated using a seed value and processed on my own server (hence the ip in the bookmarklet). Does this fit the requirements?

wa1ker38552:
I just realized that it doesn't need a bookmarklet. Here is code that you can just run by itself
```js
const width = 125;
const height = 125;

function drawAt(x, y) {
  const line = []
  const size = 1
  line.push([[x, y], [x+size, y]]);
  line.push([[x, y], [x, y+size]]);
  line.push([[x, y+size], [x+size, y+size]]);
  line.push([[x+size, y], [x+size, y+size]]);
  line.push([[x, y], [x+size, y+size]]);
  line.push([[x, y+size], [x+size, y]]);
  drawLines(line)
}

var request = new XMLHttpRequest();
request.open("GET", "https://api.codetabs.com/v1/proxy?quest=http://154.53.63.206:9000/generate", false);
request.send(null)
let graph = JSON.parse(request.responseText)

graph.forEach(function(point) {
  drawAt(point[0], point[1])
})
```

Dongathan-Jong:
Hey! Unfortunately, I cannot approve this since you are using an API to generate the art, and your code only draws said given generated art. I need to see something that you created, your code!

M0HID:
Hi there! I’m closing this PR for now to clear up the repo. Could you implement your server API code directly in the Blot editor? Once you’ve made your edits/improvements, please make a new PR and we can review it there!!

wa1ker38552:
uh, that's the problem, I can't, because I do image processing using PIL which isn't possible client side

PR #933, opened 2024-10-02
Random Dart Board By Nidhir Reddy Manne

NidhirGOAT:
This is a dart board that has 4 darts that randomly go around the dart board and can score bullseyes sometimes. It uses 51 lines of code to run the entire thing. All of the darts use the variables a through h that each two variables move the darts to a random x axis and y axis. This project took over 3 weeks to finish and a lot of editing during Hack club. The dart board is divided into 3 parts the grid is the bullseye and the vertical lines are the yellow part of the board and outside it is the outer circle of the dart board.

    - [x] Doesn't rely on color to look good

image


Dongathan-Jong:
This is a good start! Unfortunately, it falls under the bar because of the simplicity of the piece. I would recommend making it look more dart-board like, or having it on a wall and drawing out the darts too! I will be happy to re-review this when you have updated your art!

PR #945, opened 2024-10-06
Feather Quilt

DefinitelyInconspicous:
Write a short description about your art piece, such as what your artwork represents and your experience drawing it!

(Repost because old one got closed)
It was quite fun, but got a bit tedious after a while trying to solve different issues in the art. I would recommend creating more/better documentation for btoolkit as well. It was still fun though :)

Although it has some overlapping lines, Adrian tested it and it didn't rip :D

- [*] I have read the steps to [getting a blot](https://github.com/hackclub/blot/blob/main/docs/GET_A_BLOT.md)
- [*] I have read all the [submission requirements](https://blot.hackclub.com/submitting) and follow them 
- [*] I am submitting art that...
  - [*] is algorithmically generated (will meaningfully change each time the program is run)
  - [*] is your own work that is not copied from somewhere else
  - [*] is drawable through the editor(does not use outside tools)
  - [*] doesn't call `Math.random()` (See the documentation on [randomness](https://github.com/hackclub/blot/blob/main/public/TOOLKIT.md#randomness))
  - [*] is drawable on Blot itself (doesn't have lines overlap more than 5 times)
    - [*] Doesn't overlap lines more than 5 times
    - [*] Fits in the working area of 125x125mm
    - [*] Doesn't rely on colour to look good

The Blot Feather Tutorial

FeatherL


qcoral:
This is great, but it's not particularly complicated - there's a pretty big quality difference in the feather from the tutorial and the rest of the drawing. Try and close it!

Make a new PR with your changes when you're done

PR #947, opened 2024-10-06
Tree - Caden

wa1ker38552:
Generates an image of a tree. 

Snapshot (2)
Snapshot (3)
Snapshot (1)


N-coder82:
Hi @wa1ker38552, the person who merges the commits has been swamped lately, so a few of us are reviewing old PRs. This looks amazing, but the only thing I’d change is the tree. Would you be open to adding small leaves to the branches of the tree?

image

It would really bring the project to life! This is super cool, and I hope you manage to get a blot!

wa1ker38552:
I tried modifying it to do exactly that but with the rotation and angles and everything I could not get it working to a point where the leaves looked like leaves and were actually on the branch :(

saumilthecode:
just a suggestion but try replacing the colour heavy background with some artistic details like clouds/ general background things to make it look like a palm beach scene!

EnterpriseGoose:
Hello! Just wanted to note that I talked to the guy at HackClub that runs Blot and he doesn't think this tree will be able to be drawn with the blot because of how much things are overlapping. I love the shape, though, and I'd love to see you get a blot! Maybe you can figure out a new way to draw it that doesn't use the overlapping method that you're currently using. I'm excited to see what you come up with!

PR #950, opened 2024-10-08
TuxedoCupcake-Olive

EnterpriseGoose:
Original description (from PR that was closed):
This is a cupcake that's based on the cupcakes I make in real life every week. I love making a whole bunch of different flavors, so I tried to replicate that by having tons of randomization in terms of what the cake, frosting, and toppings would be. There's even a little randomization in what the cupcake wrapper looks like (which is also realistic- not every cupcake is perfect!). I struggled quite a bit trying to get the shape of the cupcake to work, and seeing as this is my first time working with something like this, it took me a bit to understand some of the documentation, and it took a lot of trial and error to get all the shapes right. Especially with the frosting curves, I tried to get both of the curve generation methods to work, but it just kept looking wrong, so I eventually found the turtle and found that it worked better at getting the arcs I was looking for. I then added a few different toppings: sprinkles, chocolate chips, and cashews (which were really hard to get drawn right). The topping choice is randomized, and then they are randomly placed on the cupcake. When it's trying to place the toppings, I originally had it completely random, but I found that it overlapped them a little too much, which I didn't think would draw well on the blot. I then added a collision detection system so they could be a little more separate, and that worked perfectly. By the end of working on this, I now feel a lot more comfortable trying to make the program actually make something the way I want it to look, and I'm excited to show this to my club so we can all work on trying to make our own drawings. I'm super excited to be able to get the blot!

Also, as much as I think it does look way cooler with the different colors, there's still some amount of randomization without them (topping placement, type of topping, base lines), and it still looks really good in black and white. I added a photo to the snapshots that shows what it looks like in black and white as well as color, which can be toggled in the program with a simple boolean.

Updates since then:
- Made it a box of cupcakes rather than just 1
- Cupcake drawing is function-driven, with custom placement and size (required refactoring most of the code)
- Color cupcake looks better (doesn't use outline color)
- Debug tools using text printing (useful in school when I don't have access to console)
- Use actual collision detection rather than just trying to guess using circles

https://github.com/EnterpriseGoose/blot-cupcake/blob/main/art/TuxedoCupcake-Olive/snapshots/gallery.png


qcoral (changes requested):
Hey!! Sorry this took so long to get to.

This is great, and I think when the color is enabled it look absolutely stunning. With it off though, it's rather static.

Try working on it more! I'll swing back around again.

EnterpriseGoose:
Hi!! Thanks so much for taking a look at this! I went through and added a few things:
- Logo on the box! This reflects how I put a little sticker on my cupcake boxes when I serve them.
- I added a cute background decoration (which can be toggled) with mini little cupcakes scattered about

I feel like there's not much more to add because there's not really any more whitespace in the space, but if you have any specific suggestions, I'd be happy to try to squeeze it in.

Thank you!!!

qcoral:
Hey!! awesome job on this.

You can order a blot here. Thanks for the patience!! https://airtable.com/appv0BzBY2APyIXj6/shrhpWVXN5imMzUmw

PR #955, opened 2024-10-09
IsometricMadness-Dragon863

Dragon863:
My submission is an easily configurable isometric cube editor. It should allow anybody with minimal code experience to be able to create their own 3d-looking art. The default image is the the letter H (for Hack Club of course!) The grid is editable using a 2D array at the start of the program, and the background lines are randomly generated and will change on each execution of the program. I have quite a bit of programming experience, but this was great for me as I've usually avoided using javascript unless necessary and this really helped me understand the language better

`(N/A)`

snapshotIsometricMadness


EnterpriseGoose:
Hello! Sorry it's been so long. The main reviewer has been swamped lately, so I wanted to give you a few hints so you have the best chance of getting a blot.
Woah, this is really cool! I love the effect you have going on here, it's super trippy!
To get a blot, though, you will need to make your project a bit more complex. Also, it will need to look good without using any fill (so having a global boolean you can change to turn off fill is necessary) Generally, you need to spend ~15-20 hrs on a project to get a blot. [See the submission guidelines here](https://blot.hackclub.com/submitting)

Some ideas I had for improvements:
- Could you make it so the isometric shapes randomize? Maybe make it in the shape of a random letter or just some other random shape. You can still have a way for you to set a custom shape, it would just add to some of the random-factor that always makes a project cool
- Although you can't use fill's for blot submissions (or, at least, it needs to look good without using fills because the blot can't draw fills), you could use lines to shade the sides of the shapes to give a similar effect
- Could you figure out how to make it 3D? Currently, you're just setting columns of effectively diagonally adjacent blocks in isometric space. If you added the ability to set blocks directly adjacent it would give the basis for a 3D effect in isometric shape. Here's a diagram that maybe can show what I mean:
image


I'm super excited to see where you go with this!

Dragon863:
Thanks so much for the feedback! My maths skills are limiting me in making it fully 3D, but I have spent a long time on adding fixes and changes; I was hoping to get a blot mostly because I thought it would be a great session to run as a workshop in our club. I've spent some time implementing several improvements, namely:
- It now looks good without using fill, and this mode can be toggled with a boolean at the top of the file
- Background density can be changed to modify the appearance of it
- Random generation is possible if the array is empty
- A few small logic errors are fixed
Here's a quick preview of the blot-drawable version (I didn't implement shading sides when on actual hardware simple because it would take forever to draw, and make the code quite unreadable)
image


Again thank you for this program, and me and my HC's members can't wait to try this out on the real thing!

PR #970, opened 2024-10-24
ninePointCircle-RishabhR

rishabhreng:
This is a visualization of the nine-point circle for any generic triangle. It is a special triangle because it contains the midpoints of each side of the triangle, the points at the feet of the altitudes, and the midpoints located between these feet and the orthocenter. I am experienced in Java for app design and robotics and Python ML but this was my first time using JS for a project, and I have learned a little about using Processing but apart from that I have not created digital art before.

https://en.wikipedia.org/wiki/Nine-point_circle

snapshot3


qcoral (changes requested):
hey - your art needs to go into the art folder!! can you update your PR with the directory changes?

rishabhreng:
Sorry about the mistake! It should be in the correct folder now.

qcoral (changes requested):
thanks for the changes!

implementation of mathematical algorithms is cool, but blot is also an artistic tool! could you think of a way to implement this with a creative element to it?

PR #975, opened 2024-10-27
Circuit_WFC-Hitu

RobbyV2:
## About

This program attempts to implement a rudimentary version of the WCSP solver, known as WFC.

Despite the initial difficulty, I was able to create a more or less working version that creates a grid with tiles, similar in looks to a PCB.

Although many of the wires are all messy, and some aren't connected, the goal was to have at least 1 path connecting all of the "chips".

It was very fun working on this, and studying on how these algorithms work!

## Image
snapshot3


qcoral (changes requested):
Hey - I really love this, but I actually have no clue what's going on. Could you add a short explanation/briefing at the top of your file?? I think it'd be super helpful for other people too. Thanks!

PR #980, opened 2024-11-01
Orbit designer-Arrow07

Arrow-07:
Write a short description about your art piece, such as what your artwork represents and your experience drawing it!

This script, through some input data and others chosen randomly, can draw the orbit of a planet around a star;

- [X ] I have read the steps to [getting a blot](https://github.com/hackclub/blot/blob/main/docs/GET_A_BLOT.md)
- [ X] I have read all the [submission requirements](https://blot.hackclub.com/submitting) and follow them 
- [ X] I am submitting art that...
  - [ X] is algorithmically generated (will meaningfully change each time the program is run)
  - [ X] is your own work that is not copied from somewhere else
  - [ X] is drawable through the editor(does not use outside tools)
  - [ X] doesn't call `Math.random()` (See the documentation on [randomness](https://github.com/hackclub/blot/blob/main/public/TOOLKIT.md#randomness))
  - [X ] is drawable on Blot itself (doesn't have lines overlap more than 5 times)
    - [X ] Doesn't overlap lines more than 5 times
    - [X ] Fits in the working area of 125x125mm
    - [X ] Doesn't rely on colour to look good

I used this guide to derive the mathematical formulas:   https://rivistaclio.com/2021/11/09/disegno-dellorbita-dei-pianeti/

snapshot1


EnterpriseGoose:
Hello! Sorry it's been so long. The main reviewer has been swamped lately, so I wanted to give you a few hints so you have the best chance of getting a blot.

Overall, I really like your project! It's really interesting technically, and it's super fun to play around with. I think you're really close to getting a blot, I would just like to see if you could add some more artistic elements.

Here are some of my ideas:
- Could you draw the star that the orbit is around and maybe a random looking planet somewhere along the calculated orbit?
- Could you add an option to calculate a random orbit?

I'm super excited to see where you go with this and I think you're really close to getting a blot!

Arrow-07:
I modified the script according to the suggestions (COMMITS OF 25/11/24), I attach the photo of how it appears now with the central star, the planet located on the orbit in a random point and the "random orbits" functionality that chooses the values ​​of x and y (therefore also of Vx and VY) randomly I spent many hours on this project and I hope to be approved :) I would be sorry to lose all this work

image

EnterpriseGoose:
The updates look great! A few more things:
- I sometimes get this error. Would you be able to fix this?
- The randomness is often creating images that are undrawable (out of bounds) or like this (which is drawable but I don't think is what you're after): 
image
could you make it so the random generation always generates something drawable?
image

I would also love to see even a bit more artistry involved. Here are a couple things I noticed:
- Could you have some kind of decoration on the planet? Maybe one that's randomly chosen or generated?
- Would it maybe make sense for the stars the planet's not orbiting around to be smaller than the main star?

It's looking great, though, and I'm excited for you to be able to get one of these last Blots of this program!

PR #991, opened 2024-11-09
jitter-scarlettekk

scarlettekk:
NOTE: because of the randomness it occasionally goes outside the bounds. Ideally the machine simply would ignore those commands. Don't know how it actually works.

I wanted to make something that represents my recent decline in mental health. I'm certainly spiraling, and the increasing amounts of jitter pretty accurately represent my less and less reliable mood and increasing anxiety and paranoia. I'm pretty happy with it, though I'd like to add some more complexity. 

I have absolutely no experience doing programming that creates visuals. I have minor python and java experience, enough to make simple modifications to existing codebases.

jitter


EnterpriseGoose:
Hello! First off, I just want to make sure you're all good :) Mental health is definitely not something to take lightly, so I would really suggest finding someone to talk to :) (from my experience art is a pretty good way of expressing yourself and getting through things, though!)

In terms of your project, I do really like the artistry of it. It definitely gives off exactly the vibes you're going for. I think you hit the right points in your pr description, though in terms of needing to make sure it doesn't draw out of bounds and increasing complexity.

In terms of making sure nothing is out of bounds, I would suggest using some bt.cut(). That was my best friend when I was making my project!

In terms of adding complexity, I have a few ideas:
- Could you have the shape of the spiral change randomly? I see you have some variables at the top there, so I'd love to see them change each run!
- Could you make the shape of the overall spiral go all wiggly? (picture so I can show what I mean:
image
(+ you would add the jitter)

I'm excited to see where you go with the project, and I hope you can get through this :)

scarlettekk:
Firstly, I'm alright, thanks for checking. I have a strong support system. I implemented your changes and tweaked the design a bit. There is now a method to the madness, with somewhat regular loops. I feel this is more accurate to my condition. There is now more randomization, the edges are trimmed, along with the style changes previously mentioned. Thank you.

EnterpriseGoose:
Hi!!! Glad you're doing well.
I like the edits; it's really fun clicking randomize and seeing all the spirals. I think Alex will want a bit more complexity, though, but to be honest I don't know exactly what would be the next thing to add, so I'll forward it off to him after these last two things:
- The name can sometimes be covered by the spiral. Could you figure out a way to keep that area clear?
- Sometime the spiral can generate in ways that are too dense (to draw on the blot), especially right in the center:
image
Is there some way you could alter the generation to account for this?

scarlettekk:
The title clipping has been resolved and the center density issue has been mostly resolved. The loops now get a little too big towards the ends, but solving that would require using some kind of exponential or power function and I need some time to brush up on my math skills (and find my TI-84). Hopefully this is an upgrade instead of a sidegrade.

PR #992, opened 2024-11-09
RandomSpirograph-FrankZ

FrankZarc:
It's a spirograph. It draws the path of a point on a circle rolling inside another circle, creating some cool patterns. 
Due to it being randomized, you can get patterns that are excessively dense and wouldn't be reasonable to draw on the actual blot. I understand if that disqualifies this.

spirograph


EnterpriseGoose:
Hello!
This is a really cool project! I love spirographs, and this is really satisfying to keep randomizing and seeing what patterns are drawn. I bet this would also be super satisfying to watch on the blot.

To get a blot, though, I would love to see a bit more complexity. Generally, people spend ~15-20 hrs on a 'blot-worthy' project, and I spent closer to 25-30.

Here are some ideas I have for making it even cooler:
- I have always loved looking at some of the weirder shaped spirographs in action. Could you program in some other shapes of spirographs? (here's a reference photo of what I'm talking about:
image
- It would be super cool if you could have it filter out images that are really boring (just a few arcs), as it would mean that you get an interesting picture every time.

In terms of drawability, I can't confirm 100% but I think this should be fine. Even the densest of the images aren't overlapping on themselves too much, and anyways it's simulating something that's supposed to be done with a pen and paper, so the blot should be able to do it just fine.

I'm super excited to see where you go with this and to try it out physically myself!

PR #998, opened 2024-11-13
AutumnTree-SmartDude1569

SmartDude1569:
This art took me weeks of effort and optimization to create. The parameters already in the file are to produce trees similar to the ones I see in my backyard. This entire thing was inspired by the trees in my backyard.

image


EnterpriseGoose:
Hello!
This is a really cool project! I love how the tree generates in such a natural looking way, and I definitely see the inspiration if I just look outside. I think you're super close to getting a blot, there are just a few things I would love to see.

- It's currently possible for leaves (ground or air) or the bottom of the trunk to be clipping out of bounds. I would suggest adjusting your algorithm and/or using bt.cut to fix this.
- Sometimes branches draw in a way where they're all overlapping and it just looks a bit like a muddled mess. Could you try to add an algorithm to the branch calculating function to just check and make sure there aren't too many things overlapping?
image
- Currently the tree is very tree looking but the leaves are not very leaf looking. Could you do something to make the leaves look a bit more like leaves? (I had to look in the code at the variable names to figure out what they were :) )
 - On that note, is there any way you could have a few leaves generate on the tree? I feel like there are always a few stragglers.

Overall I think you're super close to getting a blot, there are just those few things I think you could do to push it over the edge. I'm really excited to see where you go with this!

SmartDude1569:
thank you for this recommendation! i will make sure to make these changes as soon as possible

PR #1003, opened 2024-11-21
[Documentation Update] Update submission guide & Building instructions

EnterpriseGoose:
Summary of changes:

Building instructions
- Updated a few steps that were missing or mistyped
- Added a couple debugging tips

Submission guidelines
- Specified that not only color but the entire fill option is not drawable

PR template
- Updated to match submission guidelines

PR #1015, opened 2024-11-25
Add power switch clarification to ASSEMBLY.md

krishnans2006:
Not a new artwork; instead, it's a proposed update to the assembly instructions. I spent hours figuring out why my motors weren't working before I realized it was actually the power switch.


EnterpriseGoose:
I found this same thing! This clarification is also part of the pr I made, #1003, that covers a bunch of holes I found in the documentation.

krishnans2006:
Closed in favor of #1003.

